PR #4, opened 2016-04-25
templating attributes with mako

leifj:
This brings simple attribute templating to SATOSA.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ setup(
         "requests",
         "PyYAML",
         "gunicorn",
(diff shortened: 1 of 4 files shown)


leifj:
Please review before merging - If you like the idea, I may want to add support for getting the scope of an attribute value so you can do things like "${eduPersonAffiliation[0]@${eduPersonPrincipalName | scope}"

leifj:
Noooo - rebecka & me were still processing..,

Skickat från min iPhone

PR #6, opened 2016-04-26
template attributes... again

leifj:


rebeckag:
Applied manually together with updates in 4de5a5d.

PR #52, opened 2016-10-30
Parse encrypted assertion from IdP

skoranda:
If the SP configuration includes a key file read the key in
and use it to parse the assertion in case the assertion was
encrypted when the IdP sent it to the SP.


skoranda:
Closing to make a better merge request.

PR #53, opened 2016-10-31
Ldap attribute store

skoranda:
Adding first implementation of an LDAP attribute store microservice that users an identifier asserted by an IdP to search a LDAP directory for a person record and then asserts attributes from that person record to the SP.


skoranda:
Closing to clean up and make a more useful pull request.

PR #60, opened 2017-01-08
Per-SP configuration for LDAP attribute store microservice

skoranda:
Added functionality to the LDAP attribute store microservice
so that all details include the LDAP directory details can
be configured on a per-SP basis. Also added better logging.

### DIFF
--- a/src/satosa/micro_services/ldap_attribute_store.py
+++ b/src/satosa/micro_services/ldap_attribute_store.py
@@ -8,6 +8,7 @@ the record and assert them to the receiving SP.
 import satosa.micro_services.base
 from satosa.logging_util import satosa_logging
 
+import copy
 import logging
 import ldap3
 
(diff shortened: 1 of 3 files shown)


src/satosa/micro_services/ldap_attribute_store.py
```diff
@@ -19,88 +20,139 @@ class LdapAttributeStore(satosa.micro_services.base.ResponseMicroService):
...
+            return super().process(context, data)
+
+        satosa_logging(logger, logging.DEBUG, "{} entityID for the SP requester is {}".format(logprefix, spEntityID), context.state)
+
+        # Examine our configuration to determine if there is a per-SP configuration
+        if spEntityID in self.config:
+            config = self.config[spEntityID]
+            configClean = copy.deepcopy(config)
+            if 'bind_password' in configClean:
+                configClean['bind_password'] = 'XXXXXXXX'
```
Gijutsu: As asked above, is this line a leftover from debugging?

src/satosa/micro_services/ldap_attribute_store.py
```diff
@@ -19,88 +20,139 @@ class LdapAttributeStore(satosa.micro_services.base.ResponseMicroService):
...
 
     def process(self, context, data):
+        logprefix = LdapAttributeStore.logprefix
+
+        # Initialize the configuration to use as the default configuration
+        # that is passed during initialization.
+        config = self.config
+        configClean = copy.deepcopy(config)
+        if 'bind_password' in configClean:
+            configClean['bind_password'] = 'XXXXXXXX'
```
Gijutsu: Is this, together with the similar line at the comment below, a leftover from debugging?

src/satosa/micro_services/ldap_attribute_store.py
skoranda: No. The intent is to log the configuration at the DEBUG level but blank out any bind password included in the configuration so that passwords are never logged.

src/satosa/micro_services/ldap_attribute_store.py
skoranda: No. The intent is to log the configuration at the DEBUG level but blank out any bind password included in the configuration so that passwords are never logged.

Gijutsu: OK, thanks for the explanation, I see now how it's used while logging.

src/satosa/micro_services/ldap_attribute_store.py
Gijutsu: OK, thanks for the explanation, I see now how it's used while logging.

PR #84, opened 2017-05-09
Missing token_endpoint capabilities.

nicemid:
oic.oauth2.message.MissingRequiredAttribute: Missing required attribute
'token_endpoint'


johanlundberg:
@nicemid do you mind checking why the tests fail?

jkakavas:
This is already supported. You need to add "code" as one of the "response_types_supported" in the config and token_endpoint will be added to your configuration as seen in https://github.com/SUNET/SATOSA/blob/a7d3dff25b93c9fc8334843804b52594ae9e1c9b/src/satosa/frontends/openid_connect.py#L179-L183

jkakavas:
The test fails as it should. The test case that fails asserts that we don't advertise a token_endpoint when only the implicit grant flow is enabled.

PR #87, opened 2017-05-13
Signing signature and digest algorithm configuration

skoranda:
Enable XML signature signing and digest algorithm configuration.
A per-SAML2 IdP frontend default may be configured and per-SP
overrides may be configured. With no configuration the default
for signature algorithm is

http://www.w3.org/2001/04/xmldsig-more#rsa-sha256

and the default digist is

http://www.w3.org/2001/04/xmlenc#sha256

### DIFF
--- a/src/satosa/frontends/saml2.py
+++ b/src/satosa/frontends/saml2.py
@@ -7,7 +7,7 @@ import json
 import logging
 from urllib.parse import urlparse
 
-from saml2 import SAMLError
+from saml2 import SAMLError, xmldsig
 from saml2.config import IdPConfig
 from saml2.extension.ui import NAMESPACE as UI_NAMESPACE
 from saml2.metadata import create_metadata_string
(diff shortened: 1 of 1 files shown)


leifj:
I think this looks ok - @johanlundberg @Gijutsu @jkakavas - anyone wanna take a look before we merge?

jkakavas:
I will take a look by EOB today.

src/satosa/frontends/saml2.py
```diff
@@ -284,7 +284,46 @@ def _handle_authn_response(self, context, internal_response, idp):
...
+                'name_id'       : name_id,
+                'authn'         : auth_info,
+                'sign_response' : True
+                }
+
+        # Add the SP details
+        args.update(**resp_args)
+
+        # Default signing and digest algorithms
+        args['sign_alg'] = xmldsig.SIG_RSA_SHA256
```
jkakavas: +1 for setting sane defaults

PR #89, opened 2017-05-16
support for attribute-based authorization

leifj:
can I get some reviewz pls


leifj:
Please don't merge yet - at this point I just want review on the idea, not merge.

johanlundberg:
Maybe we want to log as well as return the auth error? Something like "Permission denied returned by AttributeAuthorization microservice" on rows 59 and 64? Would make a setup easier to debug.

johanlundberg:
@leifj Do you want to update your code to use the util function added in #99?

leifj:
just pushed the same refactor as I did for #100 to this one

PR #94, opened 2017-05-23
Attribute generation

leifj:
This microservice provides synthetic attributes using mustache templates.


johanlundberg:
This PR also contains PR #89, I think that is a mistake.

leifj:
Damn, You're right.

PR #95, opened 2017-05-24
SAMLInternalResponse for saml backend

skoranda:
Adds a SAMLInternalResponse class that the saml backend returns instead
of the parent class InternalResponse. The new class allows SAML specific
details to be added into the response that is later passed into
plugins/microservices. The first SAML specific detail added is the SAML
NameID if it is asserted by the IdP that sent the response.


leifj:
I say merged

PR #96, opened 2017-05-29
Proxy authn authority

leifj:
This sets the AuthnStatement/AuthnContext/AuthenticatingAuthority to the proxied entityID. Might be useful.


leifj:
Actually I need to create this off of a different branch...

PR #97, opened 2017-05-29
carry the proxy through as authenticationauthority

leifj:
This sets the AuthnStatement/AuthnContext/AuthenticatingAuthority to the proxied entityID. Might be useful.


leifj:
OK guys pls review!

leifj:
Yes the title is misleading and this is exactly what I'm talking about.

PR #98, opened 2017-05-29
make sign_assertion and sign_response configurable

leifj:
The saml2int profile actually requires the Assertion to be signed in addition to the response. We might want to make this configurable maybe...? however the default should be saml2int right?

### DIFF
--- a/src/satosa/frontends/saml2.py
+++ b/src/satosa/frontends/saml2.py
@@ -6,7 +6,6 @@ import functools
 import json
 import logging
 from urllib.parse import urlparse
-from distutils.util import strtobool
 
 from saml2 import SAMLError, xmldsig
 from saml2.config import IdPConfig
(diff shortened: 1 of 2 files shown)


leifj:
Yes you're right. I'll update the PR to support signed assertions but turn it off as default

leifj:
I went "nuts" and made both assertion and response signing configurable but with different defaults. My thinking is that we may need to support crazy corner-cases. I've also clarified documentation which was inconsistent with code in the use of idp_config/sp_config

src/satosa/frontends/saml2.py
```diff
@@ -6,6 +6,7 @@
 import json
 import logging
 from urllib.parse import urlparse
+from distutils.util import strtobool
```
johanlundberg: strtobool should not be needed as the config is yaml and when loaded parses True/yes/true to boolean True and False/no/false to boolean False.

skoranda:
The changes will not allow a per-SP configuration to be applied if there is no explicit default set because, for example, 

sign_assertion = self.config['idp_config']['service']['idp']['policy']['default']['sign_assertion']

will throw a KeyError and then the line

sign_assertion = self.config['idp_config']['service']['idp']['policy'][resp_args['sp_entity_id']]['sign_assertion']

will never have a chance to be evaluated.

Was that intended or should there be two (2) try/except clauses each for sign_assertion and sign_response?

johanlundberg:
Damn, I should have spotted that. I will fix it.

PR #99, opened 2017-05-30
enable defaults for custom attribute release using '' or 'default' key

leifj:
This also contains a small utility function that I will use to refactor some of my other pull requests later. There are tests covering this already so it should be safe as long as the tests work.

### DIFF
--- a/src/satosa/util.py
+++ b/src/satosa/util.py
@@ -7,6 +7,10 @@ import string
 
 logger = logging.getLogger(__name__)
 
+def get_dict_defaults(d, *keys):
+    for key in keys:
+       d = d.get(key, d.get("", d.get("default", {})))
+    return d
 
 def rndstr(size=16, alphabet=""):
     """
--- a/doc/README.md
+++ b/doc/README.md
@@ -242,6 +242,18 @@ config:
         idp-entity-id1
             sp-entity-id1:
                 exclude: ["givenName"]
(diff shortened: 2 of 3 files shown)


src/satosa/util.py
```diff
@@ -7,6 +7,10 @@
 
 logger = logging.getLogger(__name__)
 
+def get_dict_defaults(d, *keys):
+    for key in keys:
+       d = d.get(key, d.get("", d.get("default", {})))
+    return d
```
johanlundberg: Am I understanding correctly that you want to accomplish this in other words?

def get_dict_defaults(d, *keys):
    for key in keys:
        if key in d:
            return d[key]
        elif '' in d or 'default' in d:
            return d.get('', d.get('default'))
    return dict()

leifj: No. I want to traverse the entire length of the "key chain" finding the most specific branch.

leifj: Basically this is the _config function I'm using in my two other PRs for attribute authz and attribute generation. Its a  pattern for config that I think @skoranda uses too in places - esp when you need to override behaviour specified by metadata on a per-Idp, per-sp+idp or globally

johanlundberg: Ok, I get it, you want to reuse d and the keys are in hierarchical order.

PR #103, opened 2017-06-04
NameID input from attributes for LDAP attribute store

skoranda:
Add functionality to specify that attributes returned from
the LDAP directory server used with the LDAP attribute store
should be used as input to be hashed for the NameID for a
persistent name identifier sent to a SP.

### DIFF
--- a/src/satosa/micro_services/ldap_attribute_store.py
+++ b/src/satosa/micro_services/ldap_attribute_store.py
@@ -91,6 +91,12 @@ class LdapAttributeStore(satosa.micro_services.base.ResponseMicroService):
                 clear_input_attributes = self.config['clear_input_attributes']
             else:
                 clear_input_attributes = False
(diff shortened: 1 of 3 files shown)


src/satosa/micro_services/ldap_attribute_store.py
```diff
@@ -153,15 +159,41 @@ def process(self, context, data):
...
+                userId = ""
+                for attr in user_id_from_attrs:
+                    if attr in record["attributes"]:
+                        value = record["attributes"][attr]
+                        if isinstance(value, list):
+                            # Use a default sort to ensure some predictability since the
+                            # LDAP directory server may return multi-valued attributes
+                            # in any order.
+                            value.sort()
+                            for v in value:
```
jkakavas: probably nitpicking :) but `userId = "".join(value)` should be more efficient than the for loop + string concat and still be readable

PR #107, opened 2017-06-29
Sensitive config fix

leifj:


leifj:
grrr - lets try that again

PR #108, opened 2017-06-29
fix for bad config check on sensitive keys

leifj:


leifj:
sorry for the confusion - this is the right one

PR #111, opened 2017-07-11
Expose metadata endpoint via configuration option

c00kiemon5ter:
Check for configuration option `entityid_endpoint`.
When set to `true` the metadata will be served at the URL pointed by the entityid URI.

This commit introduces a new class `SAMLBaseModule` which serves as a common place for both frontend and backend _saml modules_. 

An example configuration would be:
```
name: Saml2
config:
  entityid_endpoint: true
  sp_config:
    entityid: <base_url>/<name>/proxy_saml2_backend.xml
    ...
```

### DIFF
--- a/src/satosa/backends/saml2.py
+++ b/src/satosa/backends/saml2.py
@@ -15,6 +15,7 @@ from saml2.config import SPConfig
 from saml2.extension.ui import NAMESPACE as UI_NAMESPACE
 from saml2.metadata import create_metadata_string
 
+from satosa.base import SAMLBaseModule
 from .base import BackendModule
 from ..exception import SATOSAAuthenticationError
 from ..internal_data import (InternalResponse,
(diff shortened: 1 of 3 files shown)


johanlundberg:
Even though I like the addition of the SAMLBaseModule (and think other configuration could use that abstraction) I don't know if it is ok to add it just for this config option. Or should we add it and see it as a start for similar config abstractions?

c00kiemon5ter:
The `SAMLBaseModule` class serves as a common placeholder for common functionality in both the SAML frontend and backend modules. While at the moment it only offers that option, in the future it could hold more information (I plan to do that). 

If you do not want to go forward with that, what would the alternative to share code between the frontend and backend modules be? I am open to suggestions, though I definitely do not want to duplicate the code. The only other option I can think of atm, is putting the code in a function in a generic file like _utils.py_ and calling into that (which I feel is uglier).

jkakavas:
I feel this is a cleaner way to add the common functionality needed in all SAML SATOSA plugins than any of the alternatives I can think of. And I know that @c00kiemon5ter has more functionality to submit that fits nicely on that base class so I'm all in for accepting this. We can by all means discuss it further though, if anyone has concerns

johanlundberg:
Just what I hoped you would say, that it's just a start :)

PR #112, opened 2017-07-11
Store outstanding queries to disallow unsolicited responses

c00kiemon5ter:
SATOSA does not work with `allow_unsolicited` set to `false` for a SAML backend. 

This commit introduces an in-memory dictionary that stores [outstanding queries as needed and described by pysaml2](https://github.com/rohe/pysaml2/blob/63d316554075740906282eee5967dc7149e3fd6f/src/saml2/client_base.py#L580-L582):

> A dictionary with session IDs as keys and the original web request from the user before redirection as values. 

Requests are added as new authn requests are created, and removed as matching authn responses are sent, if the module is configured not to accept unsolicited responses. If an unsolicited response is found it is logged and an `SATOSAAuthenticationError` is raised.

This could possibly be expanded to a real API that calls into some module that handles `store` and `remove` operations of the requests and gives the choice for the data to be persisted in a file or database to support across process handling of unsolicited responses.

### DIFF
--- a/src/satosa/backends/saml2.py
+++ b/src/satosa/backends/saml2.py
@@ -60,6 +60,7 @@ class SAMLBackend(BackendModule):
         self.bindings = [BINDING_HTTP_REDIRECT, BINDING_HTTP_POST]
         self.discosrv = config.get("disco_srv")
         self.encryption_keys = []
+        self.outstanding_queries = {}
 
         key_file_paths = None
         if 'encryption_keypairs' in self.config['sp_config']: # prioritize explicit encryption keypairs
@@ -120,7 +121,7 @@ class SAMLBackend(BackendModule):
         :type entity_id: str
         :rtype: satosa.response.Response
 
-        :param context: The curretn context
+        :param context: The current context
         :param entity_id: Target IDP entity id
         :return: response to the user agent
         """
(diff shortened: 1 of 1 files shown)


src/satosa/backends/saml2.py
```diff
@@ -139,6 +140,13 @@ def authn_request(self, context, entity_id):
                            exc_info=True)
             raise SATOSAAuthenticationError(context.state, "Failed to construct the AuthnRequest") from exc
 
+        if self.sp.config.allow_unsolicited is False:
```
c00kiemon5ter: oh noes! `sp.config.allow_unsolicited` does not set the correct value. 
The correct value lies under `self.sp.config._sp_allow_unsolicited`..

src/satosa/backends/saml2.py
```diff
@@ -139,6 +140,13 @@ def authn_request(self, context, entity_id):
                            exc_info=True)
             raise SATOSAAuthenticationError(context.state, "Failed to construct the AuthnRequest") from exc
 
+        if self.sp.config._sp_allow_unsolicited is False:
```
c00kiemon5ter: The PR passes the tests now, but this right here (peeking at `_`-prefixed/internal data) is really ugly. It should be fixed in `pysaml2` (`config.py`). I will have a look.

jkakavas: Can you let us know if this can be easily addressed in pysaml2 ? I agree with you that this is ugly and we shouldn't let it in

c00kiemon5ter: Update with a more standard approach.

c00kiemon5ter:
Regarding the problem that came up with the failed tests, pysaml2 `sp.config` object exposes the [`allow_unsolicited` attribute](https://github.com/rohe/pysaml2/blob/master/src/saml2/config.py#L210), which is misleading. It should be removed from the config object.

PR #120, opened 2017-07-19
Select preferred binding on the configuration

c00kiemon5ter:
`pysaml2` supports the configuration option `preferred_binding` which is a map of services to array of preferred bindings for that service. Use the configuration instead of the currently hardcoded choices.

An example configuration would be the following:

```
  name: Saml2
  config:
    sp_config:
      preferred_binding:
        single_sign_on_service:
        - urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect
        - urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST
```

This defines the preferred bindings for the SSO service, and is equivalent the code that was hardcoded before that is also the default behaviour of `pysaml2`. Changing the order of the bindings would use `HTTP-POST` instead of the `HTTP-Redirect`.

You can [see the preferred_binding service names map here](https://github.com/rohe/pysaml2/blob/master/src/saml2/config.py#L139-L156):
https://github.com/rohe/pysaml2/blob/master/src/saml2/config.py#L139-L156

### DIFF
--- a/src/satosa/backends/saml2.py
+++ b/src/satosa/backends/saml2.py
@@ -8,8 +8,6 @@ import logging
 from base64 import urlsafe_b64encode, urlsafe_b64decode
 from urllib.parse import urlparse
 
-from saml2 import BINDING_HTTP_POST
-from saml2 import BINDING_HTTP_REDIRECT
 from saml2.client_base import Base
 from saml2.config import SPConfig
 from saml2.extension.ui import NAMESPACE as UI_NAMESPACE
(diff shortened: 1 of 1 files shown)


jkakavas:
LGTM

c00kiemon5ter:
Note that this configuration option is undocumented on the `pysaml2` side. 
That should probably happen too.

jkakavas:
> Note that this configuration option is undocumented on the pysaml2 side. That should probably happen too.

Care to open an issue/PR there too ?

c00kiemon5ter:
Yes, I will

PR #121, opened 2017-07-20
Add requested authn context for backend requests

c00kiemon5ter:
Use `acr_mapping` from the backend config to generate a
`RequestedAuthnContext` node in an `AuthnRequest`.

`acr_mapping` configuration is extended so that each entry is a dictionary
with two possible keys:
- `class_ref`: defines the content of the `AuthnContextClassRef` node
- `comparison`: defines the `Comparison` attribute


c00kiemon5ter:
(closed and reopened the PR to trigger travis to run the tests again, as one failed because of network failure)

PR #124, opened 2017-07-21
Fix default values for acr

c00kiemon5ter:
The previous PR was merged too quickly to notice this. 
- Firstly, the default `Comparison` value for SAML2int should be _exact_.
- Secondly, as I discussed with @jkakavas , it does not make much sense to have a default `class_ref` value. It is the vital part of the `acr_mapping` configuration and it is mandatory that it is provided. If no value for `class_ref` is given this should fail by producing a `KeyError` exception.


johanlundberg:
Sorry, I should give some PRs more time to be discussed.

PR #125, opened 2017-07-26
Feature refactor saml modules

c00kiemon5ter:
This changeset is a generic cleanup on the saml base, frontend and backend modules. 
I tried to make the code simpler and more readable, while avoiding repetition. 

The general idea is that configuration keys should not be magic values all over the code and thus should be defined in the modules that uses the keys. Initialisation of common keys in the front and back end are handled by the base module. 

While the changes may look a lot, most are formatting changes (whitespace, max line length, etc) and there are only a couple of actual code changes on how fallbacks are handled. 
These happen on backend's `__init__` regarding `key_file_paths` and on the frontend's `_handle_authn_response` function regarding how `policy` settings are accessed to gather info about `sign_assertion`, `sign_response`, `sign_alg` and `digest_alg`.

PS: It should be easier to look at each commit separately

### DIFF
--- a/src/satosa/frontends/saml2.py
+++ b/src/satosa/frontends/saml2.py
@@ -62,6 +62,7 @@ class SAMLFrontend(FrontendModule, SAMLBaseModule):
     """
     A pysaml2 frontend module
     """
+    KEY_CUSTOM_ATTR_RELEASE = 'custom_attribute_release'
 
     def __init__(self, auth_req_callback_func, internal_attributes, config, base_url, name):
         self._validate_config(config)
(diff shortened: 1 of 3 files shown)


src/satosa/frontends/saml2.py
```diff
@@ -265,93 +276,86 @@ def _handle_authn_response(self, context, internal_response, idp):
...
+        internal_response.attributes = self._filter_attributes(
+            idp, internal_response, context)
+        ava = self.converter.from_internal(
+            self.attribute_profile, internal_response.attributes)
 
         auth_info = {}
         if self.acr_mapping:
-            auth_info["class_ref"] = self.acr_mapping.get(internal_response.auth_info.issuer, self.acr_mapping[""])
+            auth_info["class_ref"] = self.acr_mapping.get(
+                internal_response.auth_info.issuer, self.acr_mapping[""])
```
c00kiemon5ter: I think we could switch to `get_dict_defaults()` here, instead of manually trying the `""` key. 
I think this happens elsewhere too, and we can agree that the `get_dict_defaults` should dictate on how fallback keys should be checked (as a reminder there are two special/fallback keys: `""` and `"default"`).

johanlundberg: I agree.

johanlundberg:
This PR makes the code much more readable. I will merge it if no one else has something to add.

PR #128, opened 2017-07-30
Allow CMService to show additional information

jkakavas:
If a logo url/path is available for the requester ( handling happens in the respective frontend ) , send that to the consent service so that it can be shown in the UI. 
This is a non breaking change, the use case comes from the InAcademia service.

### DIFF
--- a/src/satosa/base.py
+++ b/src/satosa/base.py
@@ -103,13 +103,15 @@ class SATOSABase(object):
         state = context.state
         state[STATE_KEY] = {"requester": internal_request.requester}
         # TODO consent module should manage any state it needs by itself
-        if consent.STATE_KEY in context.state:
-            context.state[consent.STATE_KEY]["filter"] = internal_request.approved_attributes or []
-            context.state[consent.STATE_KEY]["requester_name"] = internal_request.requester_name
-        else:
-            context.state[consent.STATE_KEY] = {"filter": internal_request.approved_attributes or [],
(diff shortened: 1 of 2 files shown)


src/satosa/base.py
```diff
@@ -103,8 +103,13 @@ def _auth_req_callback_func(self, context, internal_request):
         state = context.state
         state[STATE_KEY] = {"requester": internal_request.requester}
         # TODO consent module should manage any state it needs by itself
-        context.state[consent.STATE_KEY] = {"filter": internal_request.approved_attributes or [],
-                                            "requester_name": internal_request.requester_name}
+        if consent.STATE_KEY in context.state:
```
c00kiemon5ter: What you're trying to do, is add two new values under two new keys under the `context.state[consent.STATE_KEY]` dict. What I would do is `update()` the `context.state[consent.STATE_KEY]` dict with the new key/values in one go.

```python
try:
  state_dict = context.state[consent.STATE_KEY]
except KeyError:
  state_dict = context.state[consent.STATE_KEY] = {}
finally:
  state_dict.update({
    "filter": internal_request.approved_attributes or [],
    "requester_name": internal_request.requester_name,
  })
```

So, prepare the dict (try/except), and add the values (finally/update). 
This makes the code in one block that reads what it wants to achieve and minimizes duplication.

jkakavas: Sure, thanks for the comments. I like this version better too :)

PR #130, opened 2017-08-03
Draft of error handling functionality

jkakavas:
I'm opening this so that we can discuss it. I'll close this PR once the discussion is over and submit a clean one

The idea is that one can raise a `SATOSAProcessingHaltError` from any backend/frontend or microservice and that will cause SATOSA to stop executing and either redirect the user to a given url (external error service) 
```
raise SATOSAProcessingHaltError(redirect_uri="https://this.is.url")
```

or return a 500 with a message 

```
raise SATOSAProcessingHaltError("This is an error to be shown")
```

We should probably add support for some rudimentary  templating engine ( mako ) if we want to support local errors ( 5xx, 4xx ) instead of only redirects 

Ideas and feedback are welcome

### DIFF
--- a/src/satosa/base.py
+++ b/src/satosa/base.py
@@ -8,7 +8,7 @@ from uuid import uuid4
 from satosa.micro_services import consent
 from .context import Context
 from .exception import SATOSAConfigurationError
-from .exception import SATOSAError, SATOSAAuthenticationError, SATOSAUnknownError
+from .exception import SATOSAError, SATOSAAuthenticationError, SATOSAUnknownError, SATOSAProcessingHaltError
 from .internal_data import UserIdHasher
 from .logging_util import satosa_logging
 from .micro_services.account_linking import AccountLinking
@@ -17,6 +17,7 @@ from .plugin_loader import load_backends, load_frontends
 from .plugin_loader import load_request_microservices, load_response_microservices
 from .routing import ModuleRouter, SATOSANoBoundEndpointError
 from .state import cookie_to_state, SATOSAStateError, State, state_to_cookie
+from .response import SeeOther, ServiceError
 
 
 logger = logging.getLogger(__name__)
@@ -258,6 +259,12 @@ class SATOSABase(object):
             self._save_state(resp, context)
         except SATOSANoBoundEndpointError:
             raise
+        except SATOSAProcessingHaltError as e:
+            if e.redirect_uri:
+                resp = SeeOther(redirect_url=e.redirect_uri)
+            else:
+                resp = ServiceError(message=e.message)
+
         except SATOSAError:
             satosa_logging(logger, logging.ERROR, "Uncaught SATOSA error", context.state,
                            exc_info=True)
--- a/src/satosa/exception.py
+++ b/src/satosa/exception.py
@@ -30,6 +30,39 @@ class SATOSACriticalError(SATOSAError):
     """
     pass
 
+class SATOSAProcessingHaltError(SATOSAError):
+    """
+    SATOSA error that should stop processing
+    """
+    def __init__(self, state, message, *args, **kwargs):
+        """
+        :type state: satosa.state.State
+        :param message: str
+        :param args: any
+        :param kwargs: any
+
+        :param state: Satosa state
+        :param message: A test message
+        :param args: whatever
+        :param kwargs: whatever
+        """
+        super().__init__(message, *args, **kwargs)
+        self._message = "One of the plugins or microservices raised an error. Redirect to an error page"
+        self.state = state.copy()
+        self.error_id = 0
+        if 'redirect_uri' in kwargs:
+            self.redirect_uri = kwargs['redirect_uri']
+        else:
+            self.redirect_uri = None
+
+    @property
+    def message(self):
+        """
+        :rtype: str
+        :return: Exception message
+        """
+        return self._message.format(error_id=self.error_id)
+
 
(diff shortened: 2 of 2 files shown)


src/satosa/exception.py
```diff
@@ -30,6 +30,39 @@ class SATOSACriticalError(SATOSAError):
...
+        :param args: any
+        :param kwargs: any
+
+        :param state: Satosa state
+        :param message: A test message
+        :param args: whatever
+        :param kwargs: whatever
+        """
+        super().__init__(message, *args, **kwargs)
+        self._message = "One of the plugins or microservices raised an error. Redirect to an error page"
```
c00kiemon5ter: This is a bit misleading. If this error concerns microservices only, then it should be caught [here](https://github.com/SUNET/SATOSA/blob/master/src/satosa/base.py#L174) and [here](https://github.com/SUNET/SATOSA/blob/master/src/satosa/base.py#L113), around the calls to `process()`. If this is a generic error, then I think it is placed correctly but, this message should change.

src/satosa/exception.py
```diff
@@ -30,6 +30,39 @@ class SATOSACriticalError(SATOSAError):
...
+        self.state = state.copy()
+        self.error_id = 0
+        if 'redirect_uri' in kwargs:
+            self.redirect_uri = kwargs['redirect_uri']
```
c00kiemon5ter: You can change this to just

```python
self.redirect_uri = kwargs.get('redirect_uri')
```

src/satosa/base.py
```diff
@@ -258,6 +259,12 @@ def run(self, context):
             self._save_state(resp, context)
         except SATOSANoBoundEndpointError:
             raise
+        except SATOSAProcessingHaltError as e:
+            if e.redirect_uri:
+                resp = SeeOther(redirect_url=e.redirect_uri)
+            else:
+                resp = ServiceError(message=e.message)
+
```
c00kiemon5ter: Remove this empty line and win a medal! 🥇

src/satosa/exception.py
jkakavas: What is misleading ? This is raised either in a backend/frontend or in a microservice. What makes you think it's microservice specific ?

src/satosa/exception.py
jkakavas: `kwargs.get('arg_that_doesn't_exist')` returns None ?Cool, I'll change that

src/satosa/base.py
jkakavas: Or I'll add an extra one, just for he sake of it :)

jkakavas:
Thanks for the comments @c00kiemon5ter ! Much appreciated. I would welcome some more feedback on the architecture/design/logic if anyone has some

src/satosa/exception.py
c00kiemon5ter: The place where the exception is caught is very generic. Can the core (not through a front/backend module or microservice) throw this error? 
If yes, then I think it is fine, but the message should also be more generic like `Request processing stopped due an error. Redirecting to an error page..`, as the error could originate from the core. 
If not, then the message is fine, but I would move the exception handling around the places where the error would occur (ie around the calls to `process()` for the microservices). At first I didn't understand that by _plugins_ you meant the front/backend modules, as in my head microservices are kindof like plugins. Finding and wrapping calls to front/backend modules will make things ugly, as those are scattered around the codebase.
If microservices and front/backend modules had the same interface, then this could be refactored to wrap a single call to their entry point. So, this is not something strict, just a suggestion. 
Placing the exception handling under `run()` is very convenient, but it should not be overused - in other words the following for example, is also very convenient but I would not recommend it:

```python
if __name__ == "__main__":
  try:
    main()
  except SomeVerySpecificError as e:
    log('some very specific message')
    handle_specific_error(e)
  except SomeError as e:
    log('some message')
    handle_some_error(e)
```

Errors should be caught and handled by the callers that triggered their originator. If that level is not appropriate then the exception is propagated up the stack until a suitable level - with the right contextual information for the error - can handle it (the very specific error above is most probably better handled in a deeper level). So, I would first try to find that level, than using the top level for every error. 

Now, for our case, this error is generic enough to fit this place, and I will not object having it there ;) 

But then, I am thinking, why shouldn't we redirect to an error page, for any error that is not handled deeper? ie, why should `SATOSANoBoundEndpointError`s not also be redirected to a page where the error message is shown in a nice/user-friendly way? (This would require we build a template/html-page to display errors)

jkakavas: > But then, I am thinking, why shouldn't we redirect to an error page, for any error that is not handled deeper? ie, why should SATOSANoBoundEndpointErrors not also be redirected to a page where the error message is shown in a nice/user-friendly way? (This would require we build a template/html-page to display errors) 

True that ! I wouldn't have anything against that and it fits my our InAcademia use case fine. 

Something like : 

``` python
        try:
            self._load_state(context)
            spec = self.module_router.endpoint_routing(context)
            resp = self._run_bound_endpoint(context, spec)
            self._save_state(resp, context)
        except Exception as err:
            satosa_logging(logger, logging.ERROR,
                           "Request processing stopped due to an error. Redirecting to an error page..",
                           context.state,
                           exc_info=True)
            satosa_logging(logger, logging.ERROR, err.message, context.state, exc_info=True)
            if err.redirect_uri:
                resp = SeeOther(redirect_url=err.redirect_uri)
            else:
                resp = ServiceError(message=err.message)
        return resp
```

?

PR #131, opened 2017-08-07
add scope extractor processor

jkakavas:
Processor to be used with the attribute_processor microservice. Allows to extract the scope from any scoped attribute and map that to another attribute

### DIFF
--- a/src/satosa/micro_services/processors/scope_extractor_processor.py
+++ b/src/satosa/micro_services/processors/scope_extractor_processor.py
@@ -28,7 +28,7 @@ class ScopeExtractorProcessor(BaseProcessor):
             raise AttributeProcessorError("The mapped_attribute needs to be set")
 
         attributes = internal_data.attributes
-        values = attributes.get(attribute, [None])
+        values = attributes.get(attribute, [])
         if not values:
             raise AttributeProcessorWarning("Cannot apply scope_extractor to {}, it has no values".format(attribute))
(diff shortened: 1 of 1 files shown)


src/satosa/micro_services/processors/scope_extractor_processor.py
```diff
@@ -0,0 +1,40 @@
...
+        mapped_attribute = kwargs.get(CONFIG_KEY_MAPPEDATTRIBUTE, CONFIG_DEFAULT_MAPPEDATTRIBUTE)
+        if mapped_attribute is None or mapped_attribute == '':
+            raise AttributeProcessorError("The mapped_attribute needs to be set")
+
+        attributes = internal_data.attributes
+        values = attributes.get(attribute, [None])
+        if not values:
+            raise AttributeProcessorWarning("Cannot apply scope_extractor to {}, it has no values".format(attribute))
+        if not any('@' in val for val in values):
+            raise AttributeProcessorWarning("Cannot apply scope_extractor to {}, it's values are not scoped".format(attribute))
```
jkakavas: This will not stop execution so 
```python
any('@' in val for val in values)
```
 will raise a TypeError. 
It would make more sense to return an empty list as a default value of  `get()` in 

```python
attributes.get(attribute, [None])
```

c00kiemon5ter: Sure, that would do 👍

PR #132, opened 2017-08-08
Add scope remover processor

jkakavas:
Adds a processor for the AttributeProcessor microservice that removes the scope from scoped attributes

### DIFF
--- a/src/satosa/micro_services/processors/scope_remover_processor.py
+++ b/src/satosa/micro_services/processors/scope_remover_processor.py
@@ -9,7 +9,7 @@ class ScopeRemoverProcessor(BaseProcessor):
     def process(self, internal_data, attribute, **kwargs):
         attributes = internal_data.attributes
         new_values = []
-        values = attributes.get(attribute, [None])
+        values = attributes.get(attribute, [])
         if not values:
             raise AttributeProcessorWarning("Attribute {} has no values".format(attribute))
         for value in values:
--- a/src/satosa/micro_services/processors/scope_remover_processor.py
+++ b/src/satosa/micro_services/processors/scope_remover_processor.py
@@ -1,0 +1,15 @@
+from ..attribute_processor import AttributeProcessorError
+from .base_processor import BaseProcessor
+
+class ScopeRemoverProcessor(BaseProcessor):
+    """
+    Removes the scope from a given attribute 
+    
+    """
+    def process(self, internal_data, attribute, **kwargs):
+        attributes = internal_data.attributes
+        new_values = []
+        for value in attributes.get(attribute, [None]):
+            if '@' in value:
+                new_values.append(value.split('@')[0])
+        attributes[attribute] = new_values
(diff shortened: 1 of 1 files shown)


src/satosa/micro_services/processors/scope_remover_processor.py
```diff
@@ -0,0 +1,15 @@
...
+    Removes the scope from a given attribute 
+    
+    """
+    def process(self, internal_data, attribute, **kwargs):
+        attributes = internal_data.attributes
+        new_values = []
+        for value in attributes.get(attribute, [None]):
+            if '@' in value:
+                new_values.append(value.split('@')[0])
+        attributes[attribute] = new_values
```
c00kiemon5ter: This will remove any attribute values without a `@`, and leave only values that had a scope (but with the scope-part removed). This is not clear from the class name. 

If this was not intentional you only need to remove the `if '@' in value` clause (I'm ignoring the problem described in the comment above)
```python
for value in attributes.get(attribute, [None]):
  value_with_no_scope = value.split('@')[0]
  new_values.append(value_with_no_scope)
```

src/satosa/micro_services/processors/scope_remover_processor.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ScopeRemoverProcessor(BaseProcessor):
+    """
+    Removes the scope from a given attribute 
+    
+    """
+    def process(self, internal_data, attribute, **kwargs):
+        attributes = internal_data.attributes
+        new_values = []
+        for value in attributes.get(attribute, [None]):
```
c00kiemon5ter: If `attribute` is not found, `[None]` will be returned and `value` will be set to `None`. Then `value.split()` will fail. We should probably check if the attribute is there, and then process the values. If attribute is not there, I would _raise_ an `AttributeProcessorWarning`.

src/satosa/micro_services/processors/scope_remover_processor.py
jkakavas: The idea is that the deployer uses this microservice for attributes that are scoped. I agree though that removing the attribute entirely in case it is not scoped, doesn't make much sense and will just add to confusion. I will revisit this

src/satosa/micro_services/processors/scope_remover_processor.py
jkakavas: Thanks!

src/satosa/micro_services/processors/scope_remover_processor.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Removes the scope from all values of a given attribute 
+    
+    """
+    def process(self, internal_data, attribute, **kwargs):
+        attributes = internal_data.attributes
+        new_values = []
+        values = attributes.get(attribute, [None])
+        if values:
+            for value in attributes.get(attribute, [None]):
```
c00kiemon5ter: You already have `values` set, no need to `get()` from `attributes` again ;) 

---

btw, if you want to make this "prettier", you can invert the conditional and remove the else clause and indentation

```python
if not correct:
  raise SomeErrror # processing will stop here

# "implied" else can go here now
do_stuff(...)
do_other_stuff(...)
```

This also moves the error handling closer to the cause, which makes things a bit more readable.

PR #134, opened 2017-08-11
Allow scope processor to handle multi-valued attributes

jkakavas:


src/satosa/micro_services/processors/scope_processor.py
```diff
@@ -13,5 +13,11 @@ def process(self, internal_data, attribute, **kwargs):
...
-        value = attributes.get(attribute, [None])[0]
-        attributes[attribute][0] = value + '@' + scope
+        values = attributes.get(attribute, [])
+        if not isinstance(values, list):
+            values = [values]
+        if values:
+            new_values=[]
+            for value in values:
+                new_values.append(value + '@' + scope)
+            attributes[attribute] = new_values
```
c00kiemon5ter: I would make this more concise:

```python
attributes[attribute] = list(v + '@' + scope for v in values)
```

Otherwise, LGTM

PR #135, opened 2017-08-11
Allow AL internal UniqueID to be used as attribute

jkakavas:
If configured, map AL internal unique id as an attribute to be used
by satosa, microservices and other plugins.

### DIFF
--- a/src/satosa/micro_services/account_linking.py
+++ b/src/satosa/micro_services/account_linking.py
@@ -52,6 +52,10 @@ class AccountLinking(ResponseMicroService):
             satosa_logging(logger, logging.INFO, "issuer/id pair is linked in AL service",
                            context.state)
             internal_response.user_id = message
+            id_to_attr = self.config.get("id_to_attr", None)
+            if id_to_attr:
+                internal_response.attributes[id_to_attr] = message
+
             del context.state[self.name]
             return super().process(context, internal_response)
         else:
@@ -76,6 +80,9 @@ class AccountLinking(ResponseMicroService):
             satosa_logging(logger, logging.INFO, "issuer/id pair is linked in AL service",
                            context.state)
             internal_response.user_id = message
+            id_to_attr = self.config.get("id_to_attr", None)
+            if id_to_attr:
+                internal_response.attributes[id_to_attr] = message
             try:
                 del context.state[self.name]
             except KeyError:
--- a/src/satosa/micro_services/account_linking.py
+++ b/src/satosa/micro_services/account_linking.py
@@ -32,6 +32,7 @@ class AccountLinking(ResponseMicroService):
         self.redirect_url = config["redirect_url"]
         self.signing_key = RSAKey(key=rsa_load(config["sign_key"]), use="sig", alg="RS256")
         self.endpoint = "/handle_account_linking"
+        self.config = config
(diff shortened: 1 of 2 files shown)


bajnokk:
One line was missing from the commit of @jkakavas. As this is just a trivial oversight, I prefer not to open a merge request to his tree but paste the solution here. Feel free to apply it or wait for him to update this pull request.

```diff --git a/src/satosa/micro_services/account_linking.py b/src/satosa/micro_services/account_linking.py
index ad4202e..b437129 100644
--- a/src/satosa/micro_services/account_linking.py
+++ b/src/satosa/micro_services/account_linking.py
@@ -32,6 +32,7 @@ class AccountLinking(ResponseMicroService):
         self.redirect_url = config["redirect_url"]
         self.signing_key = RSAKey(key=rsa_load(config["sign_key"]), use="sig", alg="RS256")
         self.endpoint = "/handle_account_linking"
+        self.config = config
         logger.info("Account linking is active")

     def _handle_al_response(self, context):```

File: src/satosa/micro_services/account_linking.py [hunk 1]
```diff
@@ -52,6 +53,10 @@ def _handle_al_response(self, context):
             satosa_logging(logger, logging.INFO, "issuer/id pair is linked in AL service",
                            context.state)
             internal_response.user_id = message
+            id_to_attr = self.config.get("id_to_attr", None)
```
c00kiemon5ter: `config` is static and set at initialisation time through `__init__`. I propose we move this to `__init__` too, as `id_to_attr` is not going to change and calling `.get` over and over is only wasting cpu cycles.

```diff
@@ -32,6 +32,7 @@ def __init__(self, config, *args, **kwargs):
          self.redirect_url = config["redirect_url"]
          self.signing_key = RSAKey(key=rsa_load(config["sign_key"]), use="sig", alg="RS256")
          self.endpoint = "/handle_account_linking"
+         self.config = config
+         self.id_to_attr = self.config.get("id_to_attr", None)
          logger.info("Account linking is active")
  
      def _handle_al_response(self, context):
```

then all you need to do is
```python
if self.id_to_attr:
    internal_response.attributes[self.id_to_attr] = [message]
```

src/satosa/micro_services/account_linking.py
```diff
@@ -52,6 +53,10 @@ def _handle_al_response(self, context):
...
+            if id_to_attr:
+                internal_response.attributes[id_to_attr] = message
```
c00kiemon5ter: Which brings me to the next point, which is that attributes are stored as an array of attribute-values. I think this should be

```python
internal_response.attributes[id_to_attr] = [message]
#                                          ^ notice the brackets
```

src/satosa/micro_services/account_linking.py
jkakavas: I'm only using `self.config = config` in order to get id_to_attr later on, so we even could do something like
```
self.id_to_attr = config.get("id_to_attr", None)
```

src/satosa/micro_services/account_linking.py
jkakavas: good catch, thanks

c00kiemon5ter:
LGTM

PR #137, opened 2017-08-20
add nameid-format email; add documemtation for attribute renaming

rhoerbe:
Use the frontend config entry for nameid to force a nameid-emailaddress response to the SP. (Actually used to make rocktet.chat/Meteor-SAML) compatible witht he AT-egov SAML deployment profile.

### DIFF
--- a/doc/saml-to-saml-adapter.md
+++ b/doc/saml-to-saml-adapter.md
@@ -1,0 +1,45 @@
+# saml-to-saml Adapter
+
+The might be a mismatch between the SAML profile of an IDP (1:1) or a federation (1:many) 
+and an SP with a limited SAML implementation. the proxy can convert the SAMl profile 
+to make both sides compatible. 
+
+## Support for a NameID format Emailaddress
+
+If the SP requires a NameID format EMailAddress add this configuration entry the SAMLFrontend 
+file:
+
+module: satosa.frontends.saml2.SAMLFrontend
+name: Saml2IDP
+config:
+  idp_config:
+    ...
+    service:
+      idp:
+        ...
+        name_id_format: ['urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress']
+
+## Renaming friendly attribute names
+ 
+ If an SP used friendly attribute names instead of OIDs that implementaiton should fix
+ this. In the meantime attributes can be renamed using the internal_attributes configuration. 
+ For example, to rename a backend attribute called "principalName" to "surname" at the
+ frontend, add a second attribute profile:
+ 
+    attributes:
+      ...
+      surname:
+        facebook: [last_name]
+        linkedin: [lastName]
+        openid: [family_name]
+        saml: [sn, surname]
+        saml_backend: [principalName]
+
+The additional profile 'saml_backend' needs to be registered with the backend configuration:
+
+    module: satosa.backends.saml2.SAMLBackend
+    name: Saml2
+    config:
+      attribute_profile: saml_backend
+      sp_config:
+         ...
--- a/src/satosa/frontends/saml2.py
+++ b/src/satosa/frontends/saml2.py
@@ -11,7 +11,8 @@ from saml2 import SAMLError, xmldsig
 from saml2.config import IdPConfig
 from saml2.extension.ui import NAMESPACE as UI_NAMESPACE
 from saml2.metadata import create_metadata_string
-from saml2.saml import NameID, NAMEID_FORMAT_TRANSIENT, NAMEID_FORMAT_PERSISTENT
+from saml2.saml import NameID, NAMEID_FORMAT_TRANSIENT, NAMEID_FORMAT_PERSISTENT, \
+    NAMEID_FORMAT_EMAILADDRESS
 from saml2.samlp import name_id_policy_from_string
 from saml2.server import Server
 
@@ -299,12 +300,28 @@ class SAMLFrontend(FrontendModule, SAMLBaseModule):
             attributes_to_remove = custom_release.get("exclude", [])
             for k in attributes_to_remove:
                 ava.pop(k, None)
-
-        name_id = NameID(text=internal_response.user_id,
-                         format=hash_type_to_saml_name_id_format(
-                             internal_response.user_id_hash_type),
-                         sp_name_qualifier=None,
-                         name_qualifier=None)
+        try:
+            config_nameidformat = self.config['idp_config']['service']['idp']['name_id_format'][0]
+        except KeyError:
+            config_nameidformat = None
+            satosa_logging(logger, logging.DEBUG, "No nameid format configured for frontend idp",
+                           context.state)
+        if config_nameidformat == NAMEID_FORMAT_EMAILADDRESS:
+            nameidfmt = NAMEID_FORMAT_EMAILADDRESS
+            nameidval = internal_response.attributes.get('mail', None)[0]
+            name_id = NameID(text=nameidval,
+                             format=nameidfmt,
+                             sp_name_qualifier=None,
+                             name_qualifier=None)
+        else:
+            nameidfmt = hash_type_to_saml_name_id_format(internal_response.user_id_hash_type)
+            nameidval = internal_response.user_id
+            name_id = NameID(text=nameidval,
+                             format=nameidfmt,
+                             sp_name_qualifier=None,
+                             name_qualifier=None)
+        satosa_logging(logger, logging.DEBUG, "Set nameid with format %s to '%s'" %
+                       (nameidfmt, nameidval), context.state)
 
         dbgmsg = "returning attributes %s" % json.dumps(ava)
         satosa_logging(logger, logging.DEBUG, dbgmsg, context.state)


src/satosa/frontends/saml2.py
```diff
@@ -299,12 +300,28 @@ def _handle_authn_response(self, context, internal_response, idp):
...
             for k in attributes_to_remove:
                 ava.pop(k, None)
-
-        name_id = NameID(text=internal_response.user_id,
-                         format=hash_type_to_saml_name_id_format(
-                             internal_response.user_id_hash_type),
-                         sp_name_qualifier=None,
-                         name_qualifier=None)
+        try:
+            config_nameidformat = self.config['idp_config']['service']['idp']['name_id_format'][0]
```
c00kiemon5ter: `self.idp_config` is already set, this could be 

```python
config_nameidformat = self.idp_config['service']['idp']['name_id_format'][0]
```

However, after setting `name_id` we retrieve that path again. So maybe this should be

```python
idp_conf = self.idp_config.get('service', {}).get('idp', {})
config_nameidformat = idp_conf.get('name_id_format')
if isinstance(config_nameidformat, list):
    config_nameidformat = config_nameidformat[0]
satosa_logging(logger, logging.DEBUG, "Frontend IDP nameid format: {}".format(config_nameidformat), context.state)
# ...
policies = idp_conf.get('policy', {})
```

doc/saml-to-saml-adapter.md
```diff
@@ -0,0 +1,45 @@
...
+
+module: satosa.frontends.saml2.SAMLFrontend
+name: Saml2IDP
+config:
+  idp_config:
+    ...
+    service:
+      idp:
+        ...
+        name_id_format: ['urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress']
```
c00kiemon5ter: Is `name_id_format` always a list? I think `pysaml2` [allows both a list and a value](https://github.com/rohe/pysaml2/blob/master/src/saml2/client_base.py#L318).

src/satosa/frontends/saml2.py
```diff
@@ -299,12 +300,28 @@ def _handle_authn_response(self, context, internal_response, idp):
...
+        except KeyError:
+            config_nameidformat = None
+            satosa_logging(logger, logging.DEBUG, "No nameid format configured for frontend idp",
+                           context.state)
+        if config_nameidformat == NAMEID_FORMAT_EMAILADDRESS:
+            nameidfmt = NAMEID_FORMAT_EMAILADDRESS
+            nameidval = internal_response.attributes.get('mail', None)[0]
+            name_id = NameID(text=nameidval,
```
c00kiemon5ter: `name_id` is going to be set whatever value `config_nameidformat` has. It should be moved out of the `if/else` branches.

c00kiemon5ter:
I am not sure the documentation wording is the best, but I am not the most familiar with AAI and SAML definitions, so other people should comment on this too. In particular `adapter` is not used anywhere in the project. 

There are some small typos:
```diff
- The might be a mismatch between
+ There might be a mismatch between
@@
- limited SAML implementation. the proxy
+ limited SAML implementation. The proxy 
@@
- Support for a NameID format Emailaddress
+ Support for NameID of format 'emailAddress'
@@
- If an SP used friendly attribute names instead of OIDs that implementaiton should fix this. In the meantime
+ # I think this is irrelevant to the section
@@
- add a second attribute profile
+ set different attribute_profile values for the backend and frontend
```

_Support for NameID of format `emailAddress`_ is not something that should be documented in SATOSA, but `pysaml2`. It is not a SATOSA configuration option, but an `idp` configuration and as such based on pysaml's capabilities.

Setting the `attribute_profile` configuration option is not _adding_ an attribute profile. To be able to map a protocol to itself (thus rename attributes) one needs to make sure the front and back end `attribute_profile` configuration _values_ do not clash - ie the values need to be different in order to be able to differentiate between the incoming and outgoing attributes and configure them differently in `internal_attributes.yaml` config.

doc/saml-to-saml-adapter.md
```diff
@@ -0,0 +1,45 @@
+# saml-to-saml Adapter
```
jkakavas: I'd change this to 
# Proxy: SAML2 <--> SAML2

src/satosa/frontends/saml2.py
rhoerbe: sure. (incomplete micro-refactoring:-)

doc/saml-to-saml-adapter.md
rhoerbe: ok

src/satosa/frontends/saml2.py
rhoerbe: Thanks for overlooking the code vicinity and simplifying it. I kept, however, the try/except block that has the advantage over .get() that a "missing" configuration key can be logged.

src/satosa/frontends/saml2.py
```diff
@@ -299,18 +300,36 @@ def _handle_authn_response(self, context, internal_response, idp):
...
+            config_nameidformat = idp_conf['name_id_format']
+            if isinstance(config_nameidformat, list):
+                config_nameidformat = config_nameidformat[0]
+        except KeyError:
+            config_nameidformat = None
+            satosa_logging(logger, logging.DEBUG, "No nameid format configured for frontend idp",
+                           context.state)
+        if config_nameidformat == NAMEID_FORMAT_EMAILADDRESS:
+            nameidfmt = NAMEID_FORMAT_EMAILADDRESS
+            nameidval = internal_response.attributes.get('mail', None)[0]
```
c00kiemon5ter: This will fail if `'mail'` is not set as `None` will be returned and tried to be accessed as an array (`...[0]`). It should fallback to `[None]`.

```python
nameidval = internal_response.attributes.get('mail', [None])[0]
```

rhoerbe: Happily our federation profile prescribes mail as mandatory attribute ;-)
But thanks for spotting the fault!

rhoerbe: May I ask for help?
Actually a missing mail attribute should cause an exception:

```
        if config_nameidformat == NAMEID_FORMAT_EMAILADDRESS:
            nameidfmt = NAMEID_FORMAT_EMAILADDRESS
            try:
                nameidval = internal_response.attributes['mail'][0]
            except KeyError:
                satosa_logging(logger, logging.ERROR,
                               "Missing attribute 'mail' required for nameid format emailAddress",
                               context.state)
                raise Exception("Missing attribute 'mail' required for nameid format emailAddress")
        else:
            nameidfmt = hash_type_to_saml_name_id_format(internal_response.user_id_hash_type)
            nameidval = internal_response.user_id
        name_id = NameID(text=nameidval,
                         format=nameidfmt,
                         sp_name_qualifier=None,
                         name_qualifier=None)
```
What would be the appropriate Exception? pysaml2/missingValue? A new excption?

src/satosa/frontends/saml2.py
```diff
@@ -299,18 +300,36 @@ def _handle_authn_response(self, context, internal_response, idp):
...
+            name_id = NameID(text=nameidval,
```
c00kiemon5ter: `pysaml2` is a different project (although really close to SATOSA). Exception classes defined by pysaml, should only be used by pysaml. SATOSA defines its own classes to handle exception errors and those should be used. 

If there is no matching class that describes the error well, then a new class should be defined, coupled with an appropriate error message. The class describes the family of errors, while the message describes the error. 

In this case the error stems from a misconfiguration of the provider, not SATOSA. For SATOSA this is a SAML module error (saml frontend module cannot process the response), or a requirement error (emailaddress nameid format expects the mail attribute to be set). I see no such error class defined, so you can go forward and define one. 

The exception should be raised and be chained with the underlying cause
```python
except KeyError as e:
  log(..)
  raise FamilyError("specific and descriptive error message") from e
```

---

However, there is another problem I see here. You have hardcoded the 'mail' internal attribute name. Internal attribute names are defined in the configuration and 'mail' may not be there. Internal attributes should not be hardcoded as the user specifies them -  one may prefer to name the attribute 'eMail' for example.. They are not related to the actual attribute names, they only map their values. 

What you want to do is set the `user_id_from_attrs` configuration option in `internal_attributes.yaml` to the list of internal attributes that you want to be used as the _NameID_ value (namely `user_id_from_attrs: [mail]`). Then, internally `internal_response.user_id` will match those attributes values (concatenated). 

---

So, I think this should be left unchanged. The configuration should make sure that the user_id will be generated by the _internal attribute_ that maps to the 'mail' _saml attribute_. Support for the nameid-format `emailAddress` should be added to [UserIdHashType](https://github.com/SUNET/SATOSA/blob/11edb70ef16f70022463b869c9c2219ae9100026/src/satosa/internal_data.py#L10) class.

rhoerbe: I agree, extending UserIdHashType is better.

I think that I was misled by the name 'UserIdHashType'. I understand that the class tries to generalize the NameID concept across other protocols. However, is it really limited to hashed representations as implied by the name? E.g. SAML does not prescribe to generate a persistent NameID using a hash, you could use AES or databse keys. Also, does the class intend to provide a fuzzy mapping, or should this allow for precise mapping?

Assuming that the class is not limited to hashed UserIDs it should be renamed to UserIdScopedType.

Extending the UserIdHashType enumeration with "emailAddress", together with hash_type_to_saml_name_id_format will do the job. What remains open is that this configuration option is disjointed from the frontend's idp_config/../name_form. Isname_form there just to generate metadata, or does it have any other effect?

src/satosa/internal_data.py
```diff
@@ -86,7 +87,7 @@ def hash_id(salt, user_id, requester, state):
             user_id = "{req}{time}{id}".format(req=requester, time=timestamp, id=user_id)
         elif hash_type == UserIdHashType.persistent or hash_type == UserIdHashType.pairwise:
             user_id = "{req}{id}".format(req=requester, id=user_id)
-        elif hash_type == UserIdHashType.public:
+        elif hash_type in (UserIdHashType.public, UserIdHashType.public_email):
```
c00kiemon5ter: If we're going to use this format (`x in y`) for multiple comparisons, we should do the same (in one go) a bit above, on [line 88](https://github.com/SUNET/SATOSA/pull/137/files#diff-bc07108acca2b02cc3d65ee9df483bf1R88) to keep things consistent.

src/satosa/frontends/saml2.py
```diff
@@ -56,7 +64,11 @@ def hash_type_to_saml_name_id_format(hash_type):
         return NAMEID_FORMAT_TRANSIENT
     elif hash_type == UserIdHashType.persistent.name:
         return NAMEID_FORMAT_PERSISTENT
-    return NAMEID_FORMAT_PERSISTENT
+    elif hash_type == UserIdHashType.public_email:
+        return NAMEID_FORMAT_EMAILADDRESS
+    else:
+        raise SATOSAModuleError('Mapping to SAML NameID Format {} '
+                                'not implemented'.format(hash_type.name))
```
c00kiemon5ter: This will result to the same error that #142 produces. `hash_type` may be `None`. 
I would suggest a tiny refactor (this goes for @jkakavas too) to fix this:

```
hash_name = getattr(hash_type, 'name', None)
if hash_name == ...
  ...
else:
  raise SATOSAModuleError('...{}'.format(hash_name)
```

rhoerbe:
Commit 40ca4a4 is failing - I might have drawn wrong conclusion. Must revisit this during dayligth

src/satosa/frontends/saml2.py
rhoerbe: I think that NAMEID_FORMAT_UNSPECIFIED is the appropriate mapping for hash_type None

c00kiemon5ter: I agree, this is very reasonable.

c00kiemon5ter:
I think the code is fine and also fixes #142 and #143

jkakavas:
I agree with the final approach and refactor, and with releasing transient format if we can't match ( instead of persistent failback in the original code or raising an exception )

doc/saml-to-saml-adapter.md
```diff
@@ -0,0 +1,45 @@
+# Proxy: SAML2 <--> SAML2
+
+There might be a mismatch between the SAML profile of an IDP (1:1) or a federation (1:many) 
+and an SP with a limited SAML implementation. The proxy can convert the SAMl profile 
+to make both sides compatible. 
+
+## Support for a NameID format Emailaddress
```
jkakavas: This is pysaml2 , maybe we should move this config example and section there ?

rhoerbe:
The key problem for email-NameID support is that user_id_hash_type is not set when entering _auth_resp_callback_func and _auth_resp_finish in satosa/base.py. To make my deployment working I had to remove the hashing of the userid altogether.
I appreciate any help how to fix this.

jkakavas:
@rhoerbe can you elaborate a little on the problem you mention above?

c00kiemon5ter:
The problem is that, one ends up with a _hashed value_, rather than an email-address formatted value.

---

#### `_auth_resp_callback_func()`

- collects the values of the attributes from which `user_id` is derived (configured by the `user_id_from_attrs` option)
- composes the values to derive the `user_id`
- hashes the `user_id`
- starts the microservices loop
- calls `_auth_resp_finish()`

#### `_auth_resp_finish()`

- hashes the `user_id` according to its `UserIdHashType`
- hashes any other value from the attribute defined by the `hash` configuration option
- cleans up the state

---

> What is the purpose of the first hash (call to `UserIdHasher.hash_data()` by `_auth_resp_callback_func()`)? 

I think it hashes the `user_id` so that microservices can have access to that value.

> Should the call be to `UserIdHasher.hash_data()` or `UserIdHasher.hash_id()`?

I think the intention was to use `UserIdHasher.hash_id()` as the function is there to hash the `user_id`. `UserIdHasher.hash_data()` is intended to call the hasher directly, as needed for the values defined by the `hash` configuration option.

> What's the purpose of the second hash?

as the comment says
```
# re-hash user id since e.g. account linking micro service might have changed it
```
(unless I am missing something) the value that will be hashed is either the unchanged/already-hashed value or the new/changed value. In both cases the resulting hash will be different.

---

I think the first hash should be done through `UserIdHasher.hash_id()` (as happens with the second). Once this is changed, we have a common entry point (the `hash_id()` function) where we know the `hash_type`/`UserIdHashType` and we can decide which hash-function will be invoked (call to `hash_data`) (or, we can even customize it by passing the hash-function as an argument).

When `hash_type` is `UserIdHashType.public_email` we can skip hashing altogether (the hasher is the identity function). This should solve the current situation and let the `user_id` unchanged/unhashed when the `nameid-format` is `emailAddress`.

Any comments on the above logic is appreciated.

c00kiemon5ter:
I am closing this as it has been covered by #193 and commits 75be00952e4ac9594aab796c33eb274535309f23 and d0f5552258cf027d6965303e84035984e3bba2a1. The documentation part should be moved to `pysaml2`.

PR #139, opened 2017-08-20
replace ENTRYPOINT with CMD; EXPOSE port; + apt-get clean

rhoerbe:
* CMD allows `docker run`with a shell; ENTRYPOINT is not suiteable here.- 
* For Docker orchestration tools it is good practice to export ports.
* apt-get clean at the end of each run will make the layer lean.


c00kiemon5ter:
LGTM

jkakavas:
LGTM

PR #141, opened 2017-08-23
Idp blacklisting

jkakavas:
This enables the SAML backend to define a list of blacklisted SAML IdPs that can't be used. If a user selects one of those, we return an error. 
The assumed format of the `idp_blacklist_file` is : 
```json
{"blacklist": [ "https://blacklisted_entityid1.org/saml/idp",
"https://blacklisted_entityid2.org/saml/idp",
"https://blacklisted_entityid3.org/saml/idp",
]}
```

### DIFF
--- a/example/plugins/backends/saml2_backend.yaml.example
+++ b/example/plugins/backends/saml2_backend.yaml.example
@@ -1,6 +1,7 @@
 module: satosa.backends.saml2.SAMLBackend
 name: Saml2
 config:
+  idp_blacklist_file: /path/to/blacklist.json
   sp_config:
     key_file: backend.key
     cert_file: backend.crt
@@ -13,8 +14,6 @@ config:
       local: [idp.xml]
 
     entityid: <base_url>/<name>/proxy_saml2_backend.xml
-    idp_blacklist_enabled: true
-    idp_blacklist_file: /path/to/blacklist.json
     service:
       sp:
         want_response_signed: true
--- a/example/plugins/backends/saml2_backend.yaml.example
+++ b/example/plugins/backends/saml2_backend.yaml.example
@@ -13,6 +13,8 @@ config:
       local: [idp.xml]
 
     entityid: <base_url>/<name>/proxy_saml2_backend.xml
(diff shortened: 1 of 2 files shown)


example/plugins/backends/saml2_backend.yaml.example
```diff
@@ -13,6 +13,8 @@ config:
       local: [idp.xml]
 
     entityid: <base_url>/<name>/proxy_saml2_backend.xml
+    idp_blacklist_enabled: true
```
c00kiemon5ter: Do we really need this? An undefined or empty `idp_blacklist_file` entry would suffice, imo.

src/satosa/backends/saml2.py
```diff
@@ -149,6 +149,16 @@ def authn_request(self, context, entity_id):
         :param entity_id: Target IDP entity id
         :return: response to the user agent
         """
+
+        # If IDP blacklisting is enabled and the selected IDP is blacklisted,
+        # stop here
+        if self.config["sp_config"].get("idp_blacklist_enabled", None):
```
c00kiemon5ter: The place where the setting is set, is in `pysaml2`'s configuration domain. As such it should be registered with `pysaml2`'s configuration module, and accessed as `self.sp.config.getattr("idp_blacklist_enabled", "")`. 

If we intend this to be a SATOSA configuration option (which makes sense because it enforces policy) it should be moved out of the `sp_config` block, and placed under the `config` block. Then, one can either set it on initialization (through `__init__` and use `self. idp_blacklist_enabled`) or invoke the configuration directly as `config.get("idp_blacklist_enabled")`.

example/plugins/backends/saml2_backend.yaml.example
jkakavas: yeap, could be done this way too.

src/satosa/backends/saml2.py
jkakavas: you're right, this is intended to be SATOSA config, so I'll move this to config

c00kiemon5ter:
LGTM

PR #143, opened 2017-08-25
Fix #142

### LINKED ISSUE #142: SAML Frontend always return NameID with format persistent
Function 

``` python
def hash_type_to_saml_name_id_format(hash_type):
    """
    Translate satosa format to pySAML2 name format

    :type hash_type: satosa.internal_data.UserIdHashType
    :rtype: str
    :param hash_type: satosa format
    :return: pySAML2 name format
    """
    if hash_type == UserIdHashType.transient.name:
        return NAMEID_FORMAT_TRANSIENT
    elif hash_type == UserIdHashType.persistent.name:
        return NAMEID_FORMAT_PERSISTENT
    return NAMEID_FORMAT_PERSISTENT

```

in https://github.com/SUNET/SATOSA/blob/master/src/satosa/frontends/saml2.py#L31

compares `hash_type` which is a UserIdHashType object to ` UserIdHashType.transient.name` which is a string. This comparison is always False, hence the default return is always followed and the SAML2 frontent always creates an assertion with NameID format persistent in the Subject.

jkakavas:
Make sure we compare same types

### DIFF
--- a/src/satosa/frontends/saml2.py
+++ b/src/satosa/frontends/saml2.py
@@ -52,9 +52,9 @@ def hash_type_to_saml_name_id_format(hash_type):
     :param hash_type: satosa format
     :return: pySAML2 name format
     """
-    if hash_type.name == UserIdHashType.transient.name:
+    if hash_type is UserIdHashType.transient:
         return NAMEID_FORMAT_TRANSIENT
-    elif hash_type.name == UserIdHashType.persistent.name:
+    elif hash_type is UserIdHashType.persistent:
         return NAMEID_FORMAT_PERSISTENT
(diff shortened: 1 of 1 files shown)


jkakavas:
Changed this to compare by identity, as suggested in https://docs.python.org/3/library/enum.html#comparisons

johanlundberg:
Should I merge this or wait for #137?

jkakavas:
Not sure, to be honest... I think #137 might take some more time and this is a quite significant bug (not being able to release a transient NameID ) so I say merge this and rewrite #137 based on this

johanlundberg:
Sounds good to me.

PR #144, opened 2017-08-28
Add support for reading OIDC clients from json formatted file

mrvanes:
Convenient way to pre-populate OIDC clients from readable json formatted file, instead of inserting records in MongoDB or dynamic registration. __setitem__, __delitem__ and pop are honoured but will not persist.

### DIFF
--- a/example/plugins/frontends/openid_connect_frontend.yaml.example
+++ b/example/plugins/frontends/openid_connect_frontend.yaml.example
@@ -2,7 +2,8 @@ module: satosa.frontends.openid_connect.OpenIDConnectFrontend
 name: OIDC
 config:
   signing_key_path: frontend.key
-  db_uri: mongodb://db.example.com # optional: only support MongoDB, will default to in-memory storage if not specified
+  db_uri: mongodb://db.example.com # optional: only support MongoDB, will default to in-memory storage if not specified, or
+  db_file: "metadata/oidc_clients.json" # optional: read oidc clients from json formatted file, will not persist dynamically registered clients!
   provider:
     client_registration_supported: Yes
     response_types_supported: ["code", "id_token token"]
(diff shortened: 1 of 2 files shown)


c00kiemon5ter:
Have you considered using `pickle`? 

There are other cases where data could be persisted. There could be a global `shelve` object that _pickles_ data under appropriately configured keys. That would enable SATOSA to be restarted, like no interruption occurred.

jkakavas:
[pyOP](https://github.com/SUNET/pyop), which is what we are using to implement the OIDC frontend, already supports client databases in json format that follows [OIDC client metadata](https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata). 

We probably need to make this more visible in the documentation

mrvanes:
Please enlighten me, for I only scratch our itches and didn't find anything in SATOSA, nor the pyOP lib that met our wish for simple prepopulation of OIDC clients. pyOP/storage.py only contains references to MongoDB?

jkakavas:
In SATOSA (and anything that uses pyOP), when you create a new [Provider](https://github.com/SUNET/pyop/blob/master/src/pyop/provider.py) class, you can pass a [clients](https://github.com/SUNET/pyop/blob/master/src/pyop/provider.py#L56) parameter which is a deserialized json string ( created from a json formatted string or file  with `json.load()` for instance )

The format of that json file is described in [the spec](https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata)

mrvanes:
Aha... so you suggest to stay with Mongo or memory but use the provider constructor to bootstrap with a file containing the "hardcoded" clients? How would mongowrapper react to seeing the same client twice after restart?

jkakavas:
Mark, I'm sorry but I didn't follow your thought above... We are using json formatted client db files for many projects ( with no MongoDB backend ) just fine

mrvanes:
I'll pull the pullrequest and rethink our strategy for populating the OIDC client DB, without the need for restart.

PR #145, opened 2017-08-29
set default accepted time diff, because pysaml2 would set it 0 otherwise

rhoerbe:
Not setting it makes a configuration to rare  authentication failures if clocks are synchronized well below a second.

### DIFF
--- a/example/plugins/backends/saml2_backend.yaml.example
+++ b/example/plugins/backends/saml2_backend.yaml.example
@@ -13,6 +13,7 @@ config:
       local: [idp.xml]
 
     entityid: <base_url>/<name>/proxy_saml2_backend.xml
+    accepted_time_diff: 60
     service:
       sp:
         want_response_signed: true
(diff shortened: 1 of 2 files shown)


c00kiemon5ter:
I'm OK with this, eventhough this is relevant to pysaml2. The examples provided are not meant to be complete configurations. Users must configure their services as needed. Having sane example configs is nice to have anyway.

jkakavas:
LGTM, can you make a relevant PR to the pysaml2 repo examples (optimally a PR in the documentation too ) ?

rhoerbe:
I doubt that a zero slack time is a good default in pysaml2. It shouldbe at least 5s, if not more.

Still I would say that satosa users will take the local configuration example, not that form pysaml2. Therefore the default setting should be here.

jkakavas:
>I doubt that a zero slack time is a good default in pysaml2. It shouldbe at least 5s, if not more.

>Still I would say that satosa users will take the local configuration example, not that form pysaml2. >Therefore the default setting should be here.

I didn't mean that it should be 0s. Also, I agree we should add it in the SATOSA examples but suggested we should add it to the pysaml idp/sp examples also.

PR #150, opened 2017-09-20
Use latest pyoic

johanlundberg:
Dependent on https://github.com/SUNET/pyop/pull/19.

One test is failing due to new pyoic is forcing token endpoint to be set when validation a providers configuration and the failing test checks that there should not be a token endpoint set for implicit flow.

Any insight to what is right to do here would be appreciated.

### DIFF
--- a/src/satosa/frontends/openid_connect.py
+++ b/src/satosa/frontends/openid_connect.py
@@ -73,6 +73,9 @@ class OpenIDConnectFrontend(FrontendModule):
             "scopes_supported": scopes_supported
         }
 
+        if 'code' in response_types_supported:
+            capabilities["token_endpoint"] = "{}/{}".format(endpoint_baseurl, TokenEndpoint.url)
+
         if self.config["provider"].get("client_registration_supported", False):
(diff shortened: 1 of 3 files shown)


johanlundberg:
Fixed the failing test according to @rohe's point that there is no need for a token endpoint when implicit flow is used.

alejandro-perez:
Accepting this PR is important for me, since I'm starting the implementation of support for fedoidc (in a different frontend) and it requires oidc>=0.11.0.1.

src/satosa/frontends/openid_connect.py
```diff
@@ -73,6 +73,9 @@ def _create_provider(self, endpoint_baseurl):
             "scopes_supported": scopes_supported
         }
 
+        if 'code' in response_types_supported:
+            capabilities["token_endpoint"] = "{}/jwks".format(endpoint_baseurl, TokenEndpoint.url)
```
alejandro-perez: Should be 
```
capabilities["token_endpoint"] = "{}/{}".format(endpoint_baseurl, TokenEndpoint.url)
```

PR #153, opened 2017-09-22
Fix for 151 empty internal attributes from LDAP store

skoranda:
Fix for 151 "Can internal response attributes have emptyvalues?". With
this fix the LDAP attribute store will no longer set an internal
attribute to be an empty list when the corresponding LDAP attribute
returned with the record is empty.


jkakavas:
LGTM

PR #154, opened 2017-09-22
LDAP attribute store enhancements that include breaking change for configuration syntax

skoranda:
This pull request for the LDAP attribute store microservice includes the following additions or enhancements: 

1. The ability to configure an SP to be ignored by the microservice so that no processing by the microservice is attempted. This is useful when the attributes from an IdP should be passed directly through to an SP rather than being augmented by the microservice. This is a common pattern with SPs in front of COmanage deployments.

2. Refactoring of how the attributes and NameID from an IdP are searched to find the value that should be used with the LDAP query filter. The configuration syntax has also been refactored and results in a breaking change as detailed on the satosa-dev developer's list.

3. The ability to configure a URL to which the browser will be redirected if no LDAP record is returned from the LDAP search. The default is no redirect.

It is the intention that this will be the last commit for the LDAP attribute store while it remains in the core SATOSA repository and that the microservice will soon be moved out of the core SATOSSA repository and moved into the repository for microservices and other plugins.


skoranda:
Closing this pull request since the microservices have been moved into their own repository.

PR #155, opened 2017-09-27
Updated account_linking microservice for use with eduTEAMS

jkakavas:


jkakavas:
I'll keep this here and make a duplicate one for https://github.com/IdentityPython/satosa_microservices as we need this pretty soon in eduTEAMS and we won't have the time to change configuration to handle the split of the microservices in a different repo/module

PR #156, opened 2017-09-27
Add support for static client DB configuration

### LINKED ISSUE #152: openid_connect frontend can't read clients from static file
pyOP supports reading client information from a json file but the openid_connect frontend doesn't allow for this. It only supports mongo_db or in memory dynamic client registration. 

@mrvanes captured that in https://github.com/SUNET/SATOSA/pull/144 , but I got mixed up and answered based on a combination of satosa and specific [InAcademia](https://github.com/SUNET/svs) setup.

We don't need the FileWrapper from @mrvanes PR, but the change in https://github.com/SUNET/SATOSA/pull/144/files#diff-e79c7f4af2d2a06a1b59f5b15331cffcR122 is relevant so I will make a new PR based on that

jkakavas:
As it is now it's not easy for a deployer to use SATOSA in a setup where dynamic registration is not supported and the clients are/should be statically defined. This PR allows for this and resolves #152

### DIFF
--- a/src/satosa/frontends/openid_connect.py
+++ b/src/satosa/frontends/openid_connect.py
@@ -78,11 +78,11 @@ class OpenIDConnectFrontend(FrontendModule):
 
         authz_state = self._init_authorization_state()
         db_uri = self.config.get("db_uri")
-        cdb_file = self.config.get("db_file")
+        cdb_file = self.config.get("client_db_path")
         if db_uri:
             cdb = MongoWrapper(db_uri, "satosa", "clients")
         elif cdb_file:
-            with open(self.config['client_db_path']) as f:
+            with open(cdb_file) as f:
                 cdb = json.loads(f.read())
         else:
             cdb = {}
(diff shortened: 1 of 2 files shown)


src/satosa/frontends/openid_connect.py
```diff
@@ -78,7 +78,14 @@ def _create_provider(self, endpoint_baseurl):
 
         authz_state = self._init_authorization_state()
         db_uri = self.config.get("db_uri")
-        cdb = MongoWrapper(db_uri, "satosa", "clients") if db_uri else {}
+        cdb_file = self.config.get("db_file")
```
johanlundberg: Is this line needed? I don't see where cdb_file is set.

Shouldn't the following be enough?
```python
if db_uri:
    cdb = MongoWrapper(db_uri, "satosa", "clients")
elif client_db_path:
    with open(self.config['client_db_path']) as f:
        cdb = json.loads(f.read())
else:
   cdb = {}
```

jkakavas:
My bad thanks, I'll fix this

PR #161, opened 2017-10-26
OIDC fixes

mrvanes:
While working with SATOSA for our project we bumped against multiple tiny OIDC bugs, which are conveniently cherry picked in this pull-request.
In order of appearance:
1. Only split nested attributes for oidc based profiles, this borked oid based saml attribute processing.
2. Create token_endpoint if code is supported response type, as is implied.
3. Use all incoming values for outgoing OIDC claims, may violate OIDC single valuedness of claims
4. Whitespace, automatically removed by editor.
5. Correctly extend requested OIDC sub-attributes like address.street_address

### DIFF
--- a/src/satosa/attribute_mapping.py
+++ b/src/satosa/attribute_mapping.py
@@ -95,7 +95,7 @@ class AttributeMapper(object):
 
             external_attribute_name = mapping[attribute_profile]
             attribute_values = self._collate_attribute_values_by_priority_order(external_attribute_name,
-                                                                                external_dict)
+                                                                                external_dict, attribute_profile)
             if attribute_values:  # Only insert key if it has some values
                 logger.debug("backend attribute '%s' mapped to %s" % (external_attribute_name,
                                                             internal_attribute_name))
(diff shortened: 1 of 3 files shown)


src/satosa/attribute_mapping.py
```diff
@@ -145,8 +145,11 @@ def _handle_template_attributes(self, attribute_profile, internal_dict):
 
         return internal_dict
 
-    def _get_nested_attribute_value(self, nested_key, data):
-        keys = nested_key.split(self.separator)
+    def _get_nested_attribute_value(self, nested_key, data, profile):
+        if (profile == 'openid'):
```
c00kiemon5ter: `profile` (or `attribute_profile`) is a _configuration value_ – it could be anything. I cannot be checked like that.

mrvanes: Well, as far as I understand 'openid' mapping key is fixed for all oidc frontends and backends as they appear in current codebase? See e.g. frontends/openid_connect.py @125
```
    attributes = self.converter.from_internal("openid", internal_resp.attributes)
```
All references to internal_attributes mapping dict refer to 'openid' as their mapping? It was my intention to further rely on that convention?

c00kiemon5ter: Well, that is bad. I see it hardcoded in other places too. This means that there cannot be an OIDtoOID proxy (as one would do a SAMLtoSAML proxy) – or in other words, the behaviour for openid is special. 

The `attribute_profile` should be stored at `self.attribute_profile`, and retrieved from there. I will look more into it.

baszoetekouw: Hi guys,

What's the best way going forward?  Might I suggested accepting this pull request (unless there are other pressing issues that we need to fix), and after that fixing the OpenID code path to remove hardcoded values?

mrvanes:
Superseded by PR #176

PR #165, opened 2017-10-31
pysaml2 4.4.0 -> 4.5.0

rhoerbe:
fixes broken signature validation on metadate with UTF-8 diacritics, among other things


c00kiemon5ter:
See #170

Issue #164, opened 2017-10-31
IE11 STATE_ENCRYPTION_KEY decode error 


rhoerbe:
Sign on to test.ucom.gv.at failing with "ValueError: Error 65537 while instatiating the CBC mode" when calling _load_state() in satosa.base. (Being then generalized into satosa.exception.SATOSAUnknownError)

I observed that IE11 includes and empty STATE_ENCRYPTION_KEY cookie , causing the decrypt code in cookie_to_state() to fail with an exception that is not catched in _load_state.

```
...
return CbcMode(cipher_state, iv)
File "/opt/satosa/lib/python3.5/site-packages/Cryptodome/Cipher/_mode_cbc.py", line 101, in init
% result)
ValueError: Error 65537 while instatiating the CBC mode

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
File "/opt/satosa/lib/python3.5/site-packages/satosa/proxy_server.py", line 113, in call
resp = self.run(context)
File "/opt/satosa/lib/python3.5/site-packages/satosa/base.py", line 277, in run
raise SATOSAUnknownError("Unknown error") from err
satosa.exception.SATOSAUnknownError: Unknown error
```

leifj:
This is only IE11?

rhoerbe:
We have not done cross-browser testing yet. The client is only using IE11 and FF.

leifj:
asking for a bit of review of this...

skoranda:
@rhoerbe Can you clarify if this is only observed with IE11 or if it is observed with browsers in addition to IE11? Also, can you specify a particular version or IE11? Thanks.

rhoerbe:
This was IE11 only. I tested with FF and Safari as well.

c00kiemon5ter:
I tried to reproduce this with an [IE11-vagrant-vm][vm] but couldn't. Either way, I changed the PR and I think it can be merged. If you have any comments, let me know soon; otherwise I will merge it towards the end of the week.

  [vm]: https://app.vagrantup.com/modernIE/boxes/w7-ie11

rhoerbe:
Thanks for your clean up, now this is quite shiny code :-) 

Please go ahead with the merge

PR #166, opened 2017-10-31
fix issue 164: IE11 STATE_ENCRYPTION_KEY decode error

rhoerbe:
Solution to #164  - For whatever reason decrypting the state cookie fails a WARNING message is logged and blank state initialized.

### DIFF
--- a/src/satosa/base.py
+++ b/src/satosa/base.py
@@ -225,17 +225,23 @@ class SATOSABase(object):
 
     def _load_state(self, context):
         """
-        Load a state to the context
+        Load state from cookie to the context
 
         :type context: satosa.context.Context
         :param context: Session context
         """
         try:
-            state = cookie_to_state(context.cookie, self.config["COOKIE_STATE_NAME"],
-                                    self.config["STATE_ENCRYPTION_KEY"])
-        except SATOSAStateError:
+            state = cookie_to_state(
+                    context.cookie,
+                    self.config["COOKIE_STATE_NAME"],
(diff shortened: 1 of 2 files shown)


src/satosa/base.py
```diff
@@ -222,7 +222,10 @@ def _load_state(self, context):
         try:
             state = cookie_to_state(context.cookie, self.config["COOKIE_STATE_NAME"],
                                     self.config["STATE_ENCRYPTION_KEY"])
-        except SATOSAStateError:
+        except Exception as error:
```
c00kiemon5ter: `Exception` seems too generic here, maybe revert to 

```python
except (SATOSAStateError, ValueError) as error:
  ...
```

rhoerbe: That would cause a hard failure for the client (HTTP 500 without further explanation) to the user. I thought there might be cases in the context of an AuthnRequest where a state reset might heal the problem. Combined with proper exception monitoring issues could be ironed out this way. But I have no strong feelings about this.

Gijutsu:
Can you clarify what the issue in https://github.com/IdentityPython/SATOSA/issues/164 is and if possible add a test case that triggers the problem, so that we can verify that the code change fixes it?

src/satosa/base.py
c00kiemon5ter: The error occurs within the `state.py` module. That module is responsible to handle the failure and expose it if needed as a generic `SATOSAStateError`. The error is logged and the state is reset.

PR #171, opened 2017-11-20
Full symmetric meshed metadata proxy

mrvanes:
This pullrequest makes satosa a full symmetric meshed federation hub-n-spoke saml proxy. That means that SP's can connect with mirrored IdP's and IdP's can connect with mirrored SP entityID. Meanwhile, satosa can still inject attributes or block access where or when necessary.


mrvanes:
The tests now pass the PR.

leifj:
I actually thought we had this in a microservice already

mrvanes:
I've looked, but couldn't find one that covers our requirements, which one do you mean?

mrvanes:
I've been looking at the change and could see a way to refactor get_metadata_desc() for both Front- and Backend saml2 modules. This would however require genericification of the specific sp_/idp_entities variable naming and passing a value to discern between idpsso and spsso(_descriptor). We would win brevity of the code, but completely loose readability (in my opinion) of the function. Is that really what we want?

hlflanagan:
It would be helpful if this PR was updated with the information requested in the shiny new PR template added to this project. The PR as it stands now does not have a sufficient description around how this works or why it’s built this way.

mrvanes:
New tests were added
Existing tests were modified to cater for earlier mistakes in old code
Old code mistakes have been corrected
Changes have been rebased on upstream master
Purpose of PR is creating a hub-n-spoke proxy that mimics full-meshed federation by adding the opposite of SAMLMirrorFrontend, we have extensively discussed this at TIIME
This may not be the best way to implement this, see the PR as a base to start discussing this functionality

PR #172, opened 2017-11-20
Add capability for common domain cookie

skoranda:
Add the capability to set the common domain cookie. See section
4.3.1 of
http://docs.oasis-open.org/security/saml/v2.0/saml-profiles-2.0-os.pdf.
The default remains to not set the cookie.


c00kiemon5ter:
_(posting some excerpts from the spec, for reference)_

> The name of the cookie MUST be "_saml_idp".

> The cookie value MUST be a set of one or more base-64 encoded URI values separated by a single space character
> Each URI is the unique identifier of an identity provider

> Append the identity provider's unique identifier to the list. If the identifier is already present in the list, it MAY remove and append it. The intent is that the most recently established identity provider session is the last one in the list

> The cookie MUST be set with a Path prefix of "/"
> The Domain MUST be set to ".[common-domain]". There MUST be a leading period.
> The cookie MUST be marked as secure.

> Cookie syntax should be in accordance with IETF RFC 2965
> The cookie MAY be either session-only or persistent

PR #174, opened 2017-12-15
Add metadata store SP uses to the context

skoranda:
Add the metadata store the SP is using to the context so that response
microservices have a hook to the metadata.

### DIFF
--- a/src/satosa/context.py
+++ b/src/satosa/context.py
@@ -1,6 +1,3 @@
-"""
-Holds methods for sending internal data through the satosa proxy
-"""
 from .exception import SATOSAError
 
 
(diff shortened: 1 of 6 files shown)


skoranda:
I have refactored based on discussions at the TIIME meeting in Vienna in February, 2018. Please review when you have time.

src/satosa/context.py
```diff
@@ -60,3 +61,25 @@ def path(self, p):
...
+        """
+        Add information to the context
+
+        :type context: satosa.context.Context
+        :rtype: None
+        :param context: The current context
+        :return: None
+        """
+
+        # Add the metadata store the backend is using to the context so that
```
mrvanes: I think this documention is too specific for the generalized decorator (but I get what you try to say). Maybe wrap your usecase in an 'e.g.'?

skoranda:
@mrvanes I have removed the spurious comment.

Issue #180, opened 2018-03-07
`UnboundLocalError: local variable 'attributes' referenced before assignment` in _filter_attributes (satosa/frontends/saml2.py)


xenophonf:
## Code Version
master (3.4.8)

## Expected Behavior
The `_filter_attributes` function in **satosa/frontends/saml2.py** should return all attributes or some other sensible value if no IdP policy is configured, or it should signal a human-readable error directing the user to configure a suitable IdP policy.

## Current Behavior
If no IdP policy is configured (or if it is configured incorrectly), the `_filter_attributes` function does not set a return value, which results in the following traceback:

```
[2018-03-06 23:52:06,102] [ERROR] [satosa.proxy_server]: Unknown error
Traceback (most recent call last):
  File "/opt/satosa/lib/python3.6/site-packages/satosa/base.py", line 268, in run
    resp = self._run_bound_endpoint(context, spec)
  File "/opt/satosa/lib/python3.6/site-packages/satosa/base.py", line 216, in _run_bound_endpoint
    return spec(context)
  File "/opt/satosa/lib/python3.6/site-packages/satosa/backends/saml2.py", line 235, in authn_response
    return self.auth_callback_func(context, self._translate_response(authn_response, context.state))
  File "/opt/satosa/lib/python3.6/site-packages/satosa/base.py", line 185, in _auth_resp_callback_func
    return self.response_micro_services[0].process(context, internal_response)
  File "/opt/satosa_microservices/src/satosa/micro_services/primary_identifier.py", line 203, in process
    return super().process(context, data)
  File "/opt/satosa/lib/python3.6/site-packages/satosa/micro_services/base.py", line 33, in process
    return self.next(context, data)
  File "/opt/satosa_microservices/src/satosa/micro_services/ldap_attribute_store.py", line 379, in process
    return super().process(context, data)
  File "/opt/satosa/lib/python3.6/site-packages/satosa/micro_services/base.py", line 33, in process
    return self.next(context, data)
  File "/opt/satosa/lib/python3.6/site-packages/satosa/base.py", line 159, in _auth_resp_finish
    return frontend.handle_authn_response(context, internal_response)
  File "/opt/satosa/lib/python3.6/site-packages/satosa/frontends/saml2.py", line 89, in handle_authn_response
    return self._handle_authn_response(context, internal_response, self.idp)
  File "/opt/satosa/lib/python3.6/site-packages/satosa/frontends/saml2.py", line 282, in _handle_authn_response
[...]

PR #181, opened 2018-04-09
RD - issue #180 - local variable attributes referenced before assignment

RD1991:
### All Submissions:


RD1991:
Issue #180 - 
The variable is used before assigning a value to it, this is happened due to omitting `Enable the SAML2 front end but omit the config:idp_config:service:idp:policy key from the configuration.` (as described) 

We may choose to initialise the variable with empty dictionary and then the problem would be resolved.

PR #182, opened 2018-04-24
Configurable memoization of IdP selection when using MDQ

skanct:
## The problem

When users log on one SP using their federated identity, they expect the SSO
experience, meaning that they will be able to log in to any other SP as long as
their IdP supports SSO and they have a valid authenticated session on the IdP.

SATOSA breaks this user experience as (a) the SAML frontend does not support SSO
and (b) SATOSA does not keep any state regarding the identity provider that a user has
selected. This effectively means that when SATOSA is configured to use an
external discovery service, each time a user tries to access a new SP, the user
will have to select again and again his/her identity provider, even if the user
has performed the same action just a few seconds ago. This effectively breaks
the SSO user experience.

## Discussion

There are many ways to address this behaviour. One possible route would be to
have the discovery service remember the last selected identity provider. Indeed
many DS implementations provide such a capability, but it is not without
interaction. The typical behaviour is that the user will land on the discovery
service landing page and there they will have to click on the "pre-selected" IdP
or start a new discovery if they want to change their IdP. So, effectively
the seamless SSO experience is again broken.  

The underlying reason for the existing behaviour of SATOSA is that SATOSA does
not have the notion of user sessions and does not keep user state across
different authentication flows. If we introduce user session support and state
management in SATOSA we can easily eliminate this problem and implementing full
SSO support in the SAML frontend would be trivial.

There two possible approaches that we can follow: (a) keep state on the server
side, (b) keep state on the client side. Keeping state on the server side, would
require the introduction of a session store. There is already support in SATOSA
for data stores (e.g. for monogodb), so the implementation would not be that difficult.
The downside is that in production deployments where one might have more than 1
SATOSA instances in load-balancing, high availability mode, using a shared store
requires very careful design. The other option, keeping state on the client
side, would not have this downside and SATOSA is already using a state cookie to
keep some state information in the user's browser. The downside is that we would
need to continuously transport state to and from the user's browser (which is
already the case) and that the amount information that we could store in the
cookie state would be much more limited from what we could potentially store in
a server-side data store (but the reality is that we do not need to store much
information)

We decided to go with option (b), keep state on the client side. The
implementation described below is backwards compatible, meaning that the default
behaviour of SATOSA is not altered unless the administrator uses the new
configuration options to enable the new behaviour  

## The implementation

This commit introduces four optional configuration parameters that can be used
to modify the default SATOSA behaviour:

- CONTEXT_STATE_DELETE
- remember_selected_idp_from_disco
- use_disco_when_forceauthn
- mirror_saml_forceauthn

By default, SATOSA deletes the context state when it receives an
authentication response from an identity provider. The first configuration
option, CONTEXT_STATE_DELETE, allows us disable this behaviour and thus
keeping state across different authentication flows, when the user uses
the same browser.

The second configuration option, remember_selected_idp_from_disco,
controls whether SATOSA will remember and reuse the IdP that is returned
from a discovery service. If ForceAuthn is set in the authentication
request, then the user will then the user the IdP will be redirected to
the discovery service (if it is configured) and ForceAuthn will be set
in the authentication request towards the selected IdP.

These two options together allow us to modify the current behaviour so
that within a given session, a user will select only once the identity
provider and then SATOSA will store this information in the state
cookie. When the cookie expires, the user will be redirected again to
the discovery service.

The third configuration option, use_disco_when_forceauthn, controls
whether SATOSA will redirect the use to the discovery service, even when
remember_selected_idp_from_disco is true and for the current session
there is the entity id of an IdP stored in the cookie state. This
behaviour provides a way for SPs that need to force a new IdP selection
(e.g. for account linking purposes) to use ForceAuthn to achieve this.

The fourth configuration option, mirror_saml_forceauthn, adds configuration
option to mirror ForceAuthn. By default, when the SATOSA SAML frontend
receives a SAML authentication request with ForceAuthn set to `True`, this
information is not mirrored in the SAML authentication request that is
generated by the SATOSA SAML backend towards the upstream identity provider.
If the configuration parameter `mirror_saml_forceauthn` is set to `True`,
then the default behaviour changes and the SATOSA SAML backend will set
ForceAuthn to true when it proxies a SAML authentication request with
ForceAuthn set to `True`.

The default values of these configuration options are tuned so that the
default behaviour of SATOSA is not changed.

### All Submissions:


c00kiemon5ter:
Closing this in favour of #234 which I can control and push changes to. Lets have a discussion there.

PR #191, opened 2018-08-28
Refactor target entity id

c00kiemon5ter:
This changeset fixes the misconception that the `SAMLMirrorFrontend` is leaking information to the saml2 backend. This holds only by semantics - the name of the key set in the `context` object contains `_MIRROR_`. There is no direct relation to the `SAMLMirrorFrontend`. The key could be named anything.

The changeset renames the key to a more generic form. The key is handled by the context. The context defines the _environment_ inside which the backend will run. The backend checks the environment and if the key/flag is set, it acts accordingly.

### All Submissions:

### DIFF
--- a/src/satosa/backends/saml2.py
+++ b/src/satosa/backends/saml2.py
@@ -88,18 +88,18 @@ class SAMLBackend(BackendModule, SAMLBaseModule):
         :rtype: satosa.response.Response
         """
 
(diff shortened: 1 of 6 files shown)


skoranda:
The SAMLBackend start_auth() method assumes that the target entityID passed through the context is base64 encoded. Why is that necessary?

c00kiemon5ter:
It is not and should not be necessary. I moved the decoding logic to the encoder (the SAMLMirrorFrontend).

PR #207, opened 2019-03-07
set downstream nameID

pcaskey:
Set the downstream NameID to the value of a specific attribute
regardless of what was received from the upstream IdP.

### All Submissions:

### DIFF
--- a/src/satosa/frontends/saml2.py
+++ b/src/satosa/frontends/saml2.py
@@ -332,6 +332,16 @@ class SAMLFrontend(FrontendModule, SAMLBaseModule):
         sp_policy = policies.get('default', {})
         sp_policy.update(policies.get(sp_entity_id, {}))
 
+        persistent_nameid_attr = sp_policy.get('persistent_nameid_from_attribute', {})
+        if persistent_nameid_attr:
+            satosa_logging(logger, logging.DEBUG, "Constructing persistent NameID for " + ava[persistent_nameid_attr], context.state)
+            name_id = NameID(
+                text=ava[persistent_nameid_attr],
+                format=NAMEID_FORMAT_PERSISTENT,
+                sp_name_qualifier=None,
+                name_qualifier=None,
+            )
+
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
Hello, this should be done through a plugin. Plugins/micro-services receive the internal representation of the attributes and other elements of the requests/responses. The NameID value is mapped from `internal.InternalData.subject_id` and the NameID Format from `internal.InternalData.subject_type`. Such a micro-service would look like (untested):

```python
from saml2.saml import NAMEID_FORMAT_PERSISTENT

from satosa.micro_services.base import ResponseMicroService


class NameIDFromAttribute(ResponseMicroService):
    """
    example configuration:

      module: nameid_from_attribute.NameIDFromAttribute
      name: NameIDFromAttribute
      config:
        persistent_nameid_from_attribute: eptid
    """

    def __init__(self, config, internal_attributes, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.from_attribute = config["persistent_nameid_from_attribute"]

    def process(self, context, internal_data):
        # remember, attributes can have multiple values
        # - either concat: "".join(internal_data.attributes.get(self.from_attribute, []))
        # - or select the first one: next(iter(internal_data.attributes.get(self.from_attribute)), None)
        internal_data.subject_id = "".join(internal_data.attributes.get(self.from_attribute, []))
        internal_data.subject_type = NAMEID_FORMAT_PERSISTENT
        return super().process(context, internal_data)
```

---

Another option you have is to use the [`user_id_from_attrs`](https://github.com/IdentityPython/SATOSA/tree/master/doc#user_id_from_attrs) option in `internal_attributes.yaml`. This effectively does the same as the plugin above, but I'm skeptical if it will remain or it will be removed and replaced by such a plugin.

c00kiemon5ter:
The other problematic thing with what you have is that you use a configuration option that is not declared anywhere, it is searched in the part of the configuration that belongs to `pysaml2` but it is handled by `satosa`. 
The same thing, with the configuration option, happens in the other PR #206.

PR #215, opened 2019-04-12
Add possibility to add chain for Let's Encrypt

misi:


misi:
@c00kiemon5ter 
I would be so grateful if you could review/merge this small change.

Many Thanks,
Misi

c00kiemon5ter:
This is now part of #261 with some changes

c00kiemon5ter:
merged as #278

PR #220, opened 2019-04-14
[micro-service] Set target-issuer from disco response and decide backend by target-issuer

peppelinux:
Hello,
I developed a microservice that can map specific SaToSa backends to specific target entity id. A configuration example can be this:

````
module: satosa.micro_services.custom_routing.DecideBackendByTarget
name: TargetRouter
config:
  target_mapping:
    "http://idpspid.testunical.it:8088": "spidSaml2"
    "http://strangeIDP.testunical.it:8081/saml2/metadata": "strangeSaml2"
````

In the debug logs if the microservice has been activated we can read
````
[2019-04-14 23:01:29] [DEBUG]: [urn:uuid:acedce81-9287-4700-bd96-15b7adaeab5a] Routing path: Saml2/disco
[2019-04-14 23:01:29] [DEBUG]: [urn:uuid:acedce81-9287-4700-bd96-15b7adaeab5a] Found DecideBackendByTarget (TargetRouter microservice ) redirecting Saml2 backend to spidSaml2
````

I needed a backend routing based on the target entity ID because I have some SAML2 IDP that only accepts highly customized authn request and metadata. An example could be SPID italian federation. Another example could be the need to use different configurations, like sign and digest algorithms, depending by target IDP.

I was looking into DecideBackendByRequester microservice but soon I realized that it was made for different goals, in it the subjects are the requester entity ID and not the target entity ID.

This PR also answers to my  [related issue](https://github.com/IdentityPython/SATOSA/issues/219).

### DIFF
--- a/example/plugins/microservices/target_based_routing.yaml.example
+++ b/example/plugins/microservices/target_based_routing.yaml.example
@@ -1,0 +1,12 @@
+module: satosa.micro_services.custom_routing.DecideBackendByTargetIdP
+name: TargetRouter
+config:
+  default_backend: Saml2
+
+  # the regex that will intercept http requests to be handled with this microservice
+  endpoint_paths:
+    - ".*/disco"
+
+  target_mapping:
+    "http://idpspid.testunical.it:8088": "spidSaml2" # map SAML entity with entity id 'target_id' to backend name
+    "http://eidas.testunical.it:8081/saml2/metadata": "eidasSaml2"
--- a/example/plugins/microservices/target_based_routing.yaml.example
+++ b/example/plugins/microservices/target_based_routing.yaml.example
@@ -1,11 +1,7 @@
-module: satosa.micro_services.custom_routing.DecideBackendByTargetIdP
+module: satosa.micro_services.custom_routing.DecideBackendByTargetIssuer
 name: TargetRouter
 config:
   default_backend: Saml2
-
-  # the regex that will intercept http requests to be handled with this microservice
-  endpoint_paths:
-    - ".*/disco"
 
   target_mapping:
     "http://idpspid.testunical.it:8088": "spidSaml2" # map SAML entity with entity id 'target_id' to backend name
--- a/src/satosa/micro_services/custom_routing.py
+++ b/src/satosa/micro_services/custom_routing.py
@@ -7,20 +7,17 @@ from satosa.internal import InternalData
 from .base import RequestMicroService
 from ..exception import SATOSAConfigurationError
 from ..exception import SATOSAError
-from ..exception import SATOSAStateError
 
 
 logger = logging.getLogger(__name__)
 
 
 class CustomRoutingError(SATOSAError):
-    """
-    SATOSA exception raised by CustomRouting rules
-    """
+    """SATOSA exception raised by CustomRouting rules"""
     pass
 
 
-class DecideBackendByTargetIdP(RequestMicroService):
+class DecideBackendByTargetIssuer(RequestMicroService):
     """
     Select target backend based on the target issuer.
     """
@@ -38,14 +35,11 @@ class DecideBackendByTargetIdP(RequestMicroService):
         self.default_backend = config['default_backend']
 
     def process(self, context:Context, data:InternalData):
-        """
-        Set context.target_backend based on the target issuer (context.target_entity_id)
+        """Set context.target_backend based on the target issuer"""
 
-        :param context: request context
-        :param data: the internal request
-        """
         target_issuer = context.get_decoration(Context.KEY_TARGET_ENTITYID)
         if not target_issuer:
+            logger.info('skipping backend decision because no target_issuer was found')
             return super().process(context, data)
 
         target_backend = (
@@ -62,45 +56,6 @@ class DecideBackendByTargetIdP(RequestMicroService):
 
         context.target_backend = target_backend
         return super().process(context, data)
-
-
-class DecideBackendByDiscoIdP(DecideBackendByTargetIdP):
-    def __init__(self, config:dict, *args, **kwargs):
-        super().__init__(config, *args, **kwargs)
-
-        self.disco_endpoints = config['disco_endpoints']
-        if not isinstance(self.disco_endpoints, list):
-            raise CustomRoutingError('disco_endpoints must be a list of str')
-
-    def register_endpoints(self):
-        """
-        URL mapping of additional endpoints this micro service needs to register for callbacks.
-
-        Example of a mapping from the url path '/callback' to the callback() method of a micro service:
-            reg_endp = [
-                ("^/callback1$", self.callback),
-            ]
-
-        :rtype List[Tuple[str, Callable[[satosa.context.Context, Any], satosa.response.Response]]]
-
-        :return: A list with functions and args bound to a specific endpoint url,
-                 [(regexp, Callable[[satosa.context.Context], satosa.response.Response]), ...]
-        """
-
-        return [
-            (path , self._handle_disco_response)
-            for path in self.disco_endpoints
-        ]
-
-    def _handle_disco_response(self, context:Context):
-        target_issuer_from_disco = context.request.get('entityID')
-        if not target_issuer_from_disco:
-            raise CustomRoutingError('no valid entity_id in the disco response')
-
-        context.decorate(Context.KEY_TARGET_ENTITYID, target_issuer_from_disco)
-        data_serialized = context.state.get(self.name, {}).get('internal', {})
-        data = InternalData.from_dict(data_serialized)
-        return self.process(context, data)
 
 
 class DecideBackendByRequester(RequestMicroService):
--- a/src/satosa/micro_services/custom_routing.py
+++ b/src/satosa/micro_services/custom_routing.py
@@ -2,12 +2,121 @@ import logging
 from base64 import urlsafe_b64encode
 
 from satosa.context import Context
+from satosa.internal import InternalData
+
 from .base import RequestMicroService
 from ..exception import SATOSAConfigurationError
 from ..exception import SATOSAError
+from ..exception import SATOSAStateError
 
 
 logger = logging.getLogger(__name__)
+
+
+class CustomRoutingError(SATOSAError):
+    """
+    SATOSA exception raised by CustomRouting rules
+    """
+    pass
+
+
+class DecideBackendByTargetIdP(RequestMicroService):
+    """
+    Select which backend should be used based on who is the SAML IDP
+    """
+
+    def __init__(self, config:dict, *args, **kwargs):
+        """
+        Constructor.
+        :param config: microservice configuration loaded from yaml file
+        :type config: Dict[str, Dict[str, str]]
+        """
+        super().__init__(*args, **kwargs)
+        self.target_mapping = config['target_mapping']
+        self.endpoint_paths = config['endpoint_paths']
+        self.default_backend = config['default_backend']
+
+        if not isinstance(self.endpoint_paths, list):
+            raise SATOSAConfigurationError()
+
+    def register_endpoints(self):
+        """
+        URL mapping of additional endpoints this micro service needs to register for callbacks.
+
+        Example of a mapping from the url path '/callback' to the callback() method of a micro service:
+            reg_endp = [
+                ("^/callback1$", self.callback),
+            ]
+
+        :rtype List[Tuple[str, Callable[[satosa.context.Context, Any], satosa.response.Response]]]
+
+        :return: A list with functions and args bound to a specific endpoint url,
+                 [(regexp, Callable[[satosa.context.Context], satosa.response.Response]), ...]
+        """
+
+        # this intercepts disco response
+        return [
+            (path , self.backend_by_entityid)
+            for path in self.endpoint_paths
+        ]
+
+    def _get_request_entity_id(self, context):
+        return (
+            context.get_decoration(Context.KEY_TARGET_ENTITYID) or
+            context.request.get('entityID')
+        )
+
+    def _get_backend(self, context:Context, entity_id:str) -> str:
+        """
+        returns the Target Backend to use
+        """
+        return (
+            self.target_mapping.get(entity_id) or
+            self.default_backend
+        )
+
+    def process(self, context:Context, data:dict):
+        """
+        Will modify the context.target_backend attribute based on the target entityid.
+        :param context: request context
+        :param data: the internal request
+        """
+        entity_id = self._get_request_entity_id(context)
+        if entity_id:
+            self._rewrite_context(entity_id, context)
+        return super().process(context, data)
+
+    def _rewrite_context(self, entity_id:str, context:Context) -> None:
+        tr_backend = self._get_backend(context, entity_id)
+        context.decorate(Context.KEY_TARGET_ENTITYID, entity_id)
+        context.target_frontend = context.target_frontend or context.state.get('ROUTER')
+        native_backend = context.target_backend
+        msg = (f'Found DecideBackendByTarget ({self.name} microservice) '
+               f'redirecting {entity_id} from {native_backend} '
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by peppelinux** [comment]
I force pushed all the commits to keep this contribution as more readable as possible.
In my last push I introduce this paramenter in the ms configuration:

````
  # the regex that will intercept http requests to be handled with this microservice
  endpoint_path: ".*/disco"
````

That would let users to configure many targetroutings with different endpoint registration (url path matches)

File: src/satosa/micro_services/custom_routing.py [hunk 1]
```diff
@@ -10,6 +12,66 @@
...
+
+    def backend_by_entityid(self, context):
+        entity_id = context.request.get('entityID')
+        tr_backend = self.target_mapping.get(entity_id)
+
+        if not all((context.request,
+                    entity_id,
+                    entity_id in self.target_mapping.keys(),
+                    tr_backend)):
+            return
```
c00kiemon5ter, line 58: what is the intended behaviour?

peppelinux: It's just a check, code that  I brought with me from the previous implementation.
Probably this check will return always True, so I can remove it but it's the only way to prevent exception that will block the flow (only ModuleRouting.UnknownEndpoint is handled by routing.py)

I'd like to raise ModuleRouting.UnknownEndpoint and also I'd like that UnknownEndpoint would be defined in exceptions.py. Do you agree with that?

I'll change that return with the UnknownEndpoint exception and also move this exception from ModuleRouting to exceptions.py. I can do this in this PR or in a intermediate one

c00kiemon5ter: - `context.request` should always be there
- `entity_id` may be `None`
- `tr_backend` may be `None` (which means that `entity_id in self.target_mapping.keys()` also failed)

In that case you should either have a default backend (a configuration option)
or, raise an error (don't know how to handle the request).

peppelinux: Ok, I'll raise the exception for these

peppelinux: Exception cannot be raised ... it stops the flows.
A default backend is the answer but without a constant definition, it would rely on the preexistent if available

peppelinux: that's a solution if you agree
https://github.com/IdentityPython/SATOSA/commit/d197549abc1af2fc465d23f8b3e57f4cd0bfc3f5#diff-643d6b17c79aec1dad0acafabfca8efe6dd3dcca590bcfda363aa44b06079830R54

c00kiemon5ter: > Exception cannot be raised ... it stops the flows.

Yes, it stops the flow - what else should happen if you don't know how to continue with this request?

> d197549#diff-643d6b17c79aec1dad0acafabfca8efe6dd3dcca590bcfda363aa44b06079830R54

I'm not sure what to look here. This `if` has no body.

peppelinux: Ok, completed here:

https://github.com/IdentityPython/SATOSA/commit/03afea4fb756b4582c27358a288ed6d051185cb3

thank you for the effort.
In the meanwhile, should we document this new feature? The triangle should be code, test and doc :)

peppelinux: Roger.. we have problem ... SATOSA doesn't have any documentation on MicroServices!

example/plugins/microservices/target_based_routing.yaml.example
```diff
@@ -0,0 +1,11 @@
+module: satosa.micro_services.custom_routing.DecideBackendByTarget
+name: TargetRouter
+config:
+  default_backend: Saml2
```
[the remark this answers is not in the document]
peppelinux: I think that default parameter MUST be not mandatory, in a way that a request could pass through without rewriting the context, within this microservice

src/satosa/micro_services/custom_routing.py
```diff
@@ -2,14 +2,117 @@
...
+
+    def __init__(self, config, *args, **kwargs):
+        """
+        Constructor.
+        :param config: microservice configuration loaded from yaml file
+        :type config: Dict[str, Dict[str, str]]
+        """
+        super().__init__(*args, **kwargs)
+        self.target_mapping = config['target_mapping']
+        self.endpoint_paths = config['endpoint_paths']
```
c00kiemon5ter, line 36: this says `endpoint_paths`, but the example has `endpoint_path`

src/satosa/micro_services/custom_routing.py
```diff
@@ -2,14 +2,117 @@
...
+        :param data: the internal request
+        """
+        entity_id = context.request.get('entityID')
+        if entity_id:
+            self._rewrite_context(entity_id, context)
+        return super().process(context, data)
+
+    def _rewrite_context(self, entity_id, context) -> None:
+        tr_backend = self._get_backend(context, entity_id)
+        context.internal_data['target_entity_id'] = entity_id
```
c00kiemon5ter, line 85:
```suggestion
        context.decorate(Context.KEY_TARGET_ENTITYID, entity_id)
```

src/satosa/micro_services/custom_routing.py
```diff
@@ -2,14 +2,117 @@
...
+        context.target_frontend = context.target_frontend or context.state.get('ROUTER')
```
c00kiemon5ter, line 86: why do we need to set `context.target_frontend`?

src/satosa/micro_services/custom_routing.py
peppelinux: Yes!
Fixed here https://github.com/peppelinux/SATOSA/commit/5043074ed132e7e9e8d00b4a94a41801189b6e88

src/satosa/micro_services/custom_routing.py
```diff
@@ -2,14 +2,117 @@
...
+        Constructor.
+        :param config: microservice configuration loaded from yaml file
+        :type config: Dict[str, Dict[str, str]]
+        """
+        super().__init__(*args, **kwargs)
+        self.target_mapping = config['target_mapping']
+        self.endpoint_paths = config['endpoint_paths']
+        self.default_backend = config['default_backend']
+
+        if not isinstance(self.endpoint_paths, list):
```
c00kiemon5ter, line 39: `endpoint_paths` is required to be a list, but the example shows a single value `".*/disco"`

src/satosa/micro_services/custom_routing.py
```diff
@@ -2,14 +2,117 @@
...
+        msg = (f'Found DecideBackendByTarget ({self.name} microservice) '
+               f'redirecting {entity_id} from {native_backend} '
+               f'backend to {tr_backend}')
+        logger.info(msg)
+        context.target_backend = tr_backend
+
+    def backend_by_entityid(self, context):
+        entity_id = context.request.get('entityID')
+
+        if not context.state.get('ROUTER'):
```
c00kiemon5ter, line 97: I'm guessing that here you're trying to figure out if the user accessed the endpoint directly, 
or whether it is a valid request and the user went through this endpoint as part of a flow.

I would probably create a separate properly-named function for this purposed.

src/satosa/micro_services/custom_routing.py
peppelinux: Fixed here
https://github.com/peppelinux/SATOSA/blob/DecideBackendByTarget/example/plugins/microservices/target_based_routing.yaml.example

Fixed here: https://github.com/peppelinux/SATOSA/commit/4e21146401503a3f4340fedab905de668aa7dd9f

src/satosa/micro_services/custom_routing.py
peppelinux: I used it just to have a way to know that's a flow touched by this microservice :)

c00kiemon5ter: target_frontend should already be set by the routing module. I don't think we should be changing it here.
This seems unrelated to the goal of this micro-service.

src/satosa/micro_services/custom_routing.py
```diff
@@ -2,14 +2,117 @@
...
 
 
+class CustomRoutingError(SATOSAError):
+    """
+    SATOSA exception raised by CustomRouting rules
+    """
+    pass
+
+
+class DecideBackendByTarget(RequestMicroService):
```
c00kiemon5ter, line 23: This micro-service selects
> which backend should be used based on who is the SAML IDP

This should be reflected on the name; I think we should call this `DecideBackendByTargetIdP`

src/satosa/micro_services/custom_routing.py
```diff
@@ -2,14 +2,117 @@
...
+                 [(regexp, Callable[[satosa.context.Context], satosa.response.Response]), ...]
+        """
+
+        # this intercepts disco response
+        return [
+            (path , self.backend_by_entityid)
+            for path in self.endpoint_paths
+        ]
+
+    def _get_backend(self, context:str, entity_id:str) -> str:
```
c00kiemon5ter, line 63: `context` is not a `str`, but a `satosa.context.Context` object.
It is also not used here, so we could even skip it.

```suggestion
    def _get_backend(self, context:Context, entity_id:str) -> str:
```

tests/satosa/micro_services/test_custom_routing.py
```diff
@@ -156,3 +158,77 @@ def test_missing_target_entity_id_from_context(self, context):
...
+        target_context.request = {
+            'entityID': 'unknow-entity-id',
+        }
+        target_context.state['ROUTER'] = 'Saml2'
+
+        req = InternalData(requester="test_requester")
+        assert decide_service.process(target_context, req)
+
+        res = decide_service.backend_by_entityid(target_context)
+        assert isinstance(res, InternalData)
```
c00kiemon5ter, line 221: you want to ensure that the target_backend is the expected one.
On `create_decide_service`, do
```diff
-        decide_service.next = lambda ctx, data: data
+        decide_service.next = lambda ctx, data: ctx, data
```
then
```suggestion
        expected_backend = rules['default_backend']
        res = decide_service.backend_by_entityid(target_context)
        assert ctx.get_decoration(Context.KEY_TARGET_ENTITYID) == expected_backend
```

src/satosa/micro_services/custom_routing.py
peppelinux: fixed here: https://github.com/peppelinux/SATOSA/commit/9ff7422633306646c0bb6170d5fc343026a5d9b8

src/satosa/micro_services/custom_routing.py
peppelinux: Found it, it's only required for log message, here:
https://github.com/IdentityPython/SATOSA/pull/220/files/6ccedc469751e2a441e89fec33e1f0c02a839253#diff-643d6b17c79aec1dad0acafabfca8efe6dd3dcca590bcfda363aa44b06079830R90

src/satosa/micro_services/custom_routing.py
```diff
@@ -2,14 +2,117 @@
...
+        context.target_frontend = context.target_frontend or context.state.get('ROUTER')
+        native_backend = context.target_backend
+        msg = (f'Found DecideBackendByTarget ({self.name} microservice) '
+               f'redirecting {entity_id} from {native_backend} '
+               f'backend to {tr_backend}')
+        logger.info(msg)
+        context.target_backend = tr_backend
+
+    def backend_by_entityid(self, context:Context):
+        entity_id = context.request.get('entityID')
```
c00kiemon5ter, line 95: based on the discussion on #366 
- IdpHinting should run before DecideBackendByTargetIdP
- as the name says DecideBackend "ByTargetIdP" we should be using the target-idp as this has been set by IdpHinting or other micro-services

```suggestion
        entity_id = context.get_decoration(Context.KEY_TARGET_ENTITYID)
```

src/satosa/micro_services/custom_routing.py
```diff
@@ -2,14 +2,117 @@
...
+        :param data: the internal request
+        """
+        entity_id = context.request.get('entityID')
+        if entity_id:
+            self._rewrite_context(entity_id, context)
+        return super().process(context, data)
+
+    def _rewrite_context(self, entity_id:str, context:Context) -> None:
+        tr_backend = self._get_backend(context, entity_id)
+        context.decorate(Context.KEY_TARGET_ENTITYID, entity_id)
```
c00kiemon5ter, line 85: this becomes unnecessary as that's where we've gotten entity_id from.

```suggestion
```

src/satosa/micro_services/custom_routing.py
peppelinux: done here:
dcda00627e46f56a1af7300a3131280d3405fa85#diff-643d6b17c79aec1dad0acafabfca8efe6dd3dcca590bcfda363aa44b06079830R63

src/satosa/micro_services/custom_routing.py
peppelinux: removed here
https://github.com/peppelinux/SATOSA/commit/a0a5c11af35800934d6c45c40430104aa600206b

src/satosa/micro_services/custom_routing.py
peppelinux: done here:
https://github.com/peppelinux/SATOSA/commit/dcda00627e46f56a1af7300a3131280d3405fa85#diff-643d6b17c79aec1dad0acafabfca8efe6dd3dcca590bcfda363aa44b06079830R63

c00kiemon5ter:
@peppelinux with the latest changes we now have two micro-services.

- `DiscoToTargetIssuer` registers a handler that parses a response from the discovery service and set the target_entity_id on the context.
- `DecideBackendByTargetIssuer` set the target_backend based on the target_entity_id and the mapping between issuers and backends

`DiscoToTargetIssuer` should run before `DecideBackendByTargetIssuer`.

The test are rewritten to test both services.

peppelinux:
I have to test It, probably tomorrow

peppelinux:
@ivan I MUST set `context.target_frontend` otherwise I get the following exception when the Response came back to ACS endoint

````
SATOSA/src/satosa/routing.py", line 108, in frontend_routing
    frontend = self.frontends[context.target_frontend]["instance"]
KeyError: None
````

I made my tests, It seems to me that's quite good as it is now

c00kiemon5ter:
Hello @peppelinux, I made another update, effectively moving the responsibility to set the `target_frontend` to the disco micro-service. If you were not intercepting the discovery response, there would be no reason to set the `target_frontend` on `DecideBackendByTargetIssuer`.

The reason we need to set `target_frontend`, is because it is needed to set the router state when switching from the frontend  (and the response micro-services processing) to the backend. The discovery request and response used to be handled by the backend, and at that point the router state had been set already. With the interception of the discovery response, we force some stages to run again, and those include the switch from the frontend to the backend.

Because we have navigated away from the proxy (we redirected to the discovery service) the data on context has been lost (including the target_frontend) except for the state (persisted on the cookie). Now that we need to go through the switch to the backend again, but without a target_frontend the error occurs. To fix it, we make sure that we store the target_frontend on the cookie, and retrieve it once we have a response from the discovery service.

This process does not involve the selection of the backend. Thus this is the responsibility of the `DiscoToTargetIssuer` to handle.

Ideally, `DiscoToTargetIssuer` would also create the request to the discovery service. Then this whole process would become apparent that it is needed to be done that way. We will look into doing this in the future.

c00kiemon5ter:
Please, have a look and test it. Then we should merge ;)

peppelinux:
Ok, everything works great even with idp_hinting ... Three year passed meantime but it was worth of it, let's merge it now!

peppelinux:
hey, this seems to purpose the same or a similar goal of this PR
https://gitlab.daasi.de/didmos2/didmos2-auth/-/blob/master/src/didmos_oidc/satosa/micro_services/discovery_based_routing.py

PR #221, opened 2019-04-15
Additional UnkownError informations to users

peppelinux:
A little code change to have more eloquent UnknowError message, example:

Instead of
````
Unknown error
````
we have
````
Unknown error: name 'satosa_logging' is not defined
````

### All Submissions:


c00kiemon5ter:
We will not do this. This should be part of the logs, not what the user sees.

peppelinux:
I understand. This PR introduced the same goal explained here
https://github.com/IdentityPython/SATOSA/pull/272

A human readable error sensibly decreases heldesk tickets, hope that we'll follow this goal but in a different way

PR #222, opened 2019-04-17
[Documentation] Optional multiple user_id_from_attrs

peppelinux:
Hi,
I'm using SaToSa with two Saml Backends, through [this additional feature](https://github.com/IdentityPython/SATOSA/pull/220). These two Saml2 backends gets different attributes from their IDPs (target endpoint). If one of them do not return any  'eduPersonTargetedID' but another id with a different name, then SaToSa goes in Exception `Unknow Error` because it cannot find eduPersonTargetedID as attribute name.

Using this patch I can consider multiple uid, if anyone of them could be unavailable SaToSa will try another One, following this internal_attributes example:

````
user_id_from_attrs: ["edupersontargetedid", "eppn", "thatname"]
````

I share as it came, if you will like to spend some more word on this...
thank you in advance

### All Submissions:

### DIFF
--- a/src/satosa/base.py
+++ b/src/satosa/base.py
@@ -187,7 +187,7 @@ class SATOSABase(object):
         # If configured construct the user id from attribute values.
         if "user_id_from_attrs" in self.config["INTERNAL_ATTRIBUTES"]:
             subject_id = [
-                "".join(internal_response.attributes[attr]) for attr in
+                "".join(internal_response.attributes.get(attr, '')) for attr in
                 self.config["INTERNAL_ATTRIBUTES"]["user_id_from_attrs"]
             ]
             internal_response.subject_id = "".join(subject_id)

### CODE REVIEW

**Comment by peppelinux** [comment]
@c00kiemon5ter this wouldn't introduce any breakables or configuration changements if It will be merged

**Comment by peppelinux** [comment]
As previously discussed during idpy meet, we can close this PR with an example about getting the same result with the current asset

File: src/satosa/base.py [hunk 1]
```diff
@@ -187,7 +187,7 @@ def _auth_resp_callback_func(self, context, internal_response):
         # If configured construct the user id from attribute values.
         if "user_id_from_attrs" in self.config["INTERNAL_ATTRIBUTES"]:
             subject_id = [
-                "".join(internal_response.attributes[attr]) for attr in
+                "".join(internal_response.attributes.get(attr, '')) for attr in
```
c00kiemon5ter: There are multiple things that can go wrong here. 

- If we use `.get` then `None` is a possible value, which will end up being concatenated as a string.
- Some attributes have multiple values; `attributes[attr]` could itself be a list of values.

peppelinux: I'm using this in production, of and attr Is unavailable It Will have ''. You told me about a strategy to get It to work without this patch, Just put here an example, I Will test, then we could close this PR if you agree

c00kiemon5ter:
> Using this patch I can consider multiple uid, if anyone of them could be unavailable SaToSa will try another One, following this internal_attributes example:
> 
> ```user_id_from_attrs: ["edupersontargetedid", "eppn", "thatname"]```

Instead of doing this here, you can set a new internal attribute that gets a value from all those attributes. Then this internal attribute will map to the user_id.

```
attributes:
  the_user_id:
    saml:
      - eduPersonTargetedID
      - eduPersonPrincipalName
      - mail

user_id_from_attrs:
  - the_user_id
```

---

Even more flexible would be to have a micro-service to set the `internal_data.user_id` to whatever you need, with the logic you need.

c00kiemon5ter:
Having said this, I'm not opposed to what you suggest here. I am not sure whether the semantics should be _all those attributes need to have a value_ (what currently happens) or _all those attributes may have a value_ (what you suggest). There is no way to indicate if it's one or the other, while it should be clear.

I am thinking that maybe `user_id_from_attrs` and `user_id_to_attr` should be removed entirely and `internal_attributes` should just be about the mapping.

These changes will probably come along the switch [from friendly names to canonical names](https://github.com/IdentityPython/pysaml2/issues/549).

I will leave this open for now, but we should come back to it at some point. I have some ideas on how to generalize many of these decisions so that they can be configurable, but this needs to be tried out first with something small.

peppelinux:
Great, thank you for the example, I think that's fine. If this thread could improve in any way the general asset I'll follow you

peppelinux:
Got it to work as you explained.
We could close this issue

peppelinux:
Got it to work as you explained.
We could close this issue

peppelinux:
Todo: convert this PR to Documentation, to help users to handle this use case.

PR #223, opened 2019-04-18
[satosa-saml-metadata] Load plugins from PWD

peppelinux:
This PR permit us to have a custom backed stored in working directory instead of getting the following Exception:

````
satosa-saml-metadata proxy_conf.yaml ./pki/backend.key ./pki/backend.cert --split-backend

[....]

  File "/home/user/django-saml2-idp.env/lib/python3.7/site-packages/satosa/plugin_loader.py", line 195, in _load_plugin_module
    raise ValueError("Can't find module '%s'" % plugin_config["module"])
ValueError: Can't find module 'backends.spidsaml2.SpidSAMLBackend'
````

I have a custom backend called SpidSAMLBackend in a local folder, where the project configuration is.
Adding this simple line of code now I can get it to work

### All Submissions:

### DIFF
--- a/src/satosa/plugin_loader.py
+++ b/src/satosa/plugin_loader.py
@@ -190,6 +190,7 @@ def _load_endpoint_module(plugin_config, plugin_filter):
 
 
 def _load_plugin_module(plugin_config, plugin_filter):
+    sys.path.append('.') # permits to load plugins from PWD
     module_class = locate(plugin_config["module"])
     if not module_class:
         raise ValueError("Can't find module '%s'" % plugin_config["module"])


c00kiemon5ter:
We will be moving away from this type of loading. This assumes that all micro-services are under the respective dir of the project. Instead, we will split all frontends, backends and micro-services to be their own packages. We will only look at the system python-path to load modules.
This is actually what is already happening but we will offload to `importlib` all the machinery needed to handle this.

If you want to load that backend you have to make it available in the python path yourself. The easiest way is to have a basic `setup.py` file and `pip install -e .` in that project. So, a virtualenv can see both satosa and that project, and thus when satosa is run, this project will be available to load.

peppelinux:
Got It

PR #233, opened 2019-05-20
Add functionality to assert static SAML attributes per CO

skoranda:
Added functionality to the frontends.saml2.SAMLVirtualCoFrontend class
so that the frontend SAML IdP can be configured to assert a set of SAML
attributes with static values for all users per CO.

### All Submissions:

### DIFF
--- a/example/plugins/frontends/saml2_virtualcofrontend.yaml.example
+++ b/example/plugins/frontends/saml2_virtualcofrontend.yaml.example
@@ -24,6 +24,15 @@ config:
         - contact_type: technical
           email_address: help@messproject.org
           given_name MESS Technical Support
+       # SAML attributes and static values about the CO to be asserted for each user.
+       # The key is the SATOSA internal attribute name.
+       co_static_saml_attributes:
+         organization: Medium Engergy Synchrotron Source
+         countryname: US
+         friendlycountryname: United States
+         noreduorgacronym:
+           - MESS
+           - MeSyncS
     - encodeable_name: MTS
       organization:
         display_name: Milwaukee Theological Seminary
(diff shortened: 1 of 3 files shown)


example/plugins/frontends/saml2_virtualcofrontend.yaml.example
```diff
@@ -24,6 +24,15 @@ config:
...
           given_name MESS Technical Support
+       # SAML attributes and static values about the CO to be asserted for each user.
+       # The key is the SATOSA internal attribute name.
+       co_static_saml_attributes:
+         organization: Medium Engergy Synchrotron Source
+         countryname: US
+         friendlycountryname: United States
+         noreduorgacronym:
+           - MESS
+           - MeSyncS
```
c00kiemon5ter: This block seems to have one space more in indentation.
It looks like it is under `contact_person` but it shouldn't.

src/satosa/frontends/saml2.py
```diff
@@ -726,11 +727,25 @@ def handle_authn_response(self, context, internal_response):
...
         idp = self._create_co_virtual_idp(context)
-        return self._handle_authn_response(context, internal_response, idp)
+
+        # Add any static attributes for the CO.
+        co_config = self._get_co_config(context)
+
+        if self.KEY_CO_ATTRIBUTES in co_config:
+            attributes = internal_response.attributes
+            for attribute, value in co_config[self.KEY_CO_ATTRIBUTES].items():
+                attributes[attribute] = value
```
c00kiemon5ter: Internal-attribute values are of type `list`; this is because some attributes can have multiple values.
If you want to keep the same interface, I'd make this 

```
from typing import Iterable

[...]

attributes[attribute] = (
    value                          # keep the value as is
    if isinstance(value, Iterable) # if it is already a list (like noreduorgacronym)
    and not isinstance(value, str) # and not a string (strings are iterables)
    else [value]                   # otherwise wrap it in a list
)
```

skoranda:
Thanks @c00kiemon5ter, I fixed the issues and made sure tests again passed.

skoranda:
@c00kiemon5ter This fix works with Python 3.4. I included a comment to refactor when Python 3.4 is no longer supported.

PR #234, opened 2019-05-27
Feature memorize idp

c00kiemon5ter:
Based on #182 

- rebased against current master
- renamed configuration options
- cleaned up to be easier to understand

### All Submissions:


tests/satosa/backends/test_saml2.py
```diff
@@ -226,32 +242,6 @@ def test_authn_request(self, context, idp_conf):
         req_params = dict(parse_qsl(urlparse(resp.message).query))
         assert context.state[self.samlbackend.name]["relay_state"] == req_params["RelayState"]
 
-    def test_mirror_saml_forceauthn(self, context, sp_conf):
```
c00kiemon5ter: This test needs to be restored. It can now be done through `get_force_authn` function.

doc/README.md
```diff
@@ -322,6 +323,59 @@ config:
...
+
+The default behaviour is `False`.
+
+```yaml
+config:
+  mirror_force_authn: True
+  [...]
+```
+
+##### Memorize the IdP selected through the discovery service
```
c00kiemon5ter: This has now changed, and happens regardless of whether the selection came from the discovery service.
There is actually, no other selection mechanism other than the discovery service. In all other cases there is either only one IdP (no choice), or the IdP has been chose as part of the mirrored-frontend (see `Context.KEY_TARGET_ENTITYID` - again, no choice).

doc/README.md
```diff
@@ -322,6 +323,59 @@ config:
...
+
+The default behaviour is `False`.
+
+```yaml
+config:
+  memorize_idp: True
+  [...]
+```
+
+##### Use the configured discovery service if ForceAuthn is set to true
```
c00kiemon5ter: Using `use_memorized_idp_when_force_authn` allows backwards compatibility as its absence means it is set to `False`. This was previously the option `use_disco_when_forceauthn` which had the invert logic - it needed the default value `True` which feels non-intuitive.

src/satosa/backends/saml2.py
```diff
@@ -35,6 +35,31 @@
 logger = logging.getLogger(__name__)
 
 
+def get_memorized_idp(context, config, force_authn):
```
c00kiemon5ter: Tests for the config options can now be written against this function.

Tests for the preference order of the different entity-id options can be written against the `get_idp_entity_id()` function.

src/satosa/backends/saml2.py
```diff
@@ -35,6 +35,31 @@
...
+        and context.state.get(Context.KEY_MEMORIZED_IDP)
+    )
+    use_when_force_authn = config.get(
+        SAMLBackend.KEY_USE_MEMORIZED_IDP_WHEN_FORCE_AUTHN
+    )
+    value = (not force_authn or use_when_force_authn) and memorized_idp
+    return value
+
+
+# XXX check KEY_FORCE_AUTHN value type (boolean vs str)
```
c00kiemon5ter: We definitely need to do that. The problem here is that this value will be store/serialized in `context.state` when the user will be redirected to the discovery service. Since JSON has no boolean type, this will be serialized as string and needs to be converted back (`False` will be serialized as "False" which is a _truthy_ value).

One option here is to actually use numbers:
```python
In [1]: int(True)                                                                                                                                                                                               
Out[1]: 1

In [2]: int(False)                                                                                                                                                                                              
Out[2]: 0

In [3]: bool(int(True))                                                                                                                                                                                         
Out[3]: True

In [4]: bool(int(False))                                                                                                                                                                                        
Out[4]: False
```

src/satosa/backends/saml2.py
```diff
@@ -35,6 +35,31 @@
...
+def get_force_authn(context, config, sp_config):
+    value = (
+        config.get(SAMLBackend.KEY_MIRROR_FORCE_AUTHN)
```
c00kiemon5ter: Only take into account `Context.KEY_FORCE_AUTHN` if `SAMLBackend.KEY_MIRROR_FORCE_AUTHN` has been set.

src/satosa/backends/saml2.py
```diff
@@ -35,6 +35,31 @@
...
+        and (
+            context.state.get(Context.KEY_FORCE_AUTHN)
+            or context.get_decoration(Context.KEY_FORCE_AUTHN)
+        )
+        or sp_config.getattr("force_authn", "sp")
```
c00kiemon5ter: `force_authn` might be set to `True` through the _pysaml2 configuration_ for this SP. We need to take that into account when deciding if we're going to redirect to the discovery service or use the memorized-idp (when set)

src/satosa/backends/saml2.py
```diff
@@ -85,26 +116,60 @@ def __init__(self, outgoing, internal_attributes, config, base_url, name):
...
-        # if there is only one IdP in the metadata, bypass the discovery service
-        idps = self.sp.metadata.identity_providers()
-        if len(idps) == 1 and "mdq" not in self.config["sp_config"]["metadata"]:
-            entity_id = idps[0]
-            return self.authn_request(context, entity_id)
+        entity_id = self.get_idp_entity_id(context)
+        if entity_id is None:
+            # since context is not passed to disco_query
+            # keep the information in the state cookie
+            context.state[Context.KEY_FORCE_AUTHN] = get_force_authn(
```
c00kiemon5ter: `State` should probably provide methods to (de)serialize given values.

src/satosa/backends/saml2.py
```diff
@@ -248,8 +320,11 @@ def authn_response(self, context, binding):
             raise SATOSAAuthenticationError(context.state, "State did not match relay state")
 
         context.decorate(Context.KEY_BACKEND_METADATA_STORE, self.sp.metadata)
-
-        del context.state[self.name]
+        if self.config.get(SAMLBackend.KEY_MEMORIZE_IDP):
```
c00kiemon5ter: No need to actually check for this - it does not matter how we store, but how/when we use it.

src/satosa/backends/saml2.py
```diff
@@ -181,7 +249,11 @@ def authn_request(self, context, entity_id):
         kwargs = {}
         authn_context = self.construct_requested_authn_context(entity_id)
         if authn_context:
-            kwargs['requested_authn_context'] = authn_context
+            kwargs["requested_authn_context"] = authn_context
+        if self.config.get(SAMLBackend.KEY_MIRROR_FORCE_AUTHN):
```
c00kiemon5ter: No need to actually check for this; we could always be setting `force_authn` to the returned value by `get_force_authn()`

src/satosa/backends/saml2.py
```diff
@@ -248,8 +320,11 @@ def authn_response(self, context, binding):
             raise SATOSAAuthenticationError(context.state, "State did not match relay state")
 
         context.decorate(Context.KEY_BACKEND_METADATA_STORE, self.sp.metadata)
-
-        del context.state[self.name]
+        if self.config.get(SAMLBackend.KEY_MEMORIZE_IDP):
+            issuer = authn_response.response.issuer.text.strip()
+            context.state[Context.KEY_MEMORIZED_IDP] = issuer
```
c00kiemon5ter: At this point we know that the authn was successful and we store the issuer/IdP. Previously, this was done without checking the response (see bcd238893a53f1d66ac3c2083f18a1239155790d), but at that point we knew that the IdP was the one selected from the discovery service. We know lose this information, but it does not seem to be very valuable - see https://github.com/IdentityPython/SATOSA/pull/234/files#r287894650

PR #237, opened 2019-06-07
Granular logging control

rhoerbe:
Setting loglevel to debug produces a quite voluminous output, because various levels of
requests are traced. This PR aims for providing more control per module/function.

This PR uses the logging Filter objects instead of wrapping log.debug() etc.
While it is necessary to get correct function names in the filter, 
it has also the advantage to keep the logging.log signature.

As a consequence, satosa_logging needs to be replaced such as in this example:

satosa_logging(logger, logging.ERROR, "Could not find necessary info about entity: %s" % e, context.state)

-> 

logger.error("Could not find necessary info about entity: %s" % e, extra={'state': context.state})

In the first round I converted backend.saml2, frontend.saml2, proxy_server and state.

Remarks: 
(1) The configuration allows to skip or shorten messages, and to process them with caller-defined functions.
See satosa_log_filter.py.

(2) The configuration is stored as a module-level dict in satosa_log_filter.py. 
I.e., the configuration dict is initialized with the first import and cached for every other import.
This approach might be better than the curren passing around. 
I had a case where it would have been be useful to access the proxy configuration in a microservice (not the module config only).


c00kiemon5ter:
The logs should always be complete. How one will process the logs is not something that concerns the project. How logs will be (post-)processed is left upon the infrastructure of each user to decide, using external tools.

Different people will have different needs and we cannot satisfy all of those. We are currently allowing for customization of the python logger through the [`LOGGING`](https://github.com/IdentityPython/SATOSA/blob/71e114d/example/proxy_conf.yaml.example#L17-L43) configuration option which allows for setting the format to include the module, file and function name. 
When we get to revamping the whole logging infra of this project this might also go away. (The plan is to move towards structured logging. Each log-line will be a single JSON-formatted line, with lots of information that will be bound to the logger (pid, thread-id, hostname, version, env, module & file info, etc), plus the given payload or exception trace.

rhoerbe:
"The logs should always be complete" - Yes, for INFO level and upwards. However, in production settings it would be uncool to flood an ELK-stack with lots of DEBUG data, just to drill into a specific issue.

My other considerations was that I want a comparable functionality to Shibboleth's logging control out of the box. `LOGGING` is limiting control to severity and format. In particular the frequent tracebacks are a pain for any post-processing.

The trade-off for more (optional) control is just an import and adding a log filter per module, which is quite lean. (And refactoring the satosa_logging wrapper, which is a minor change.)

I see no conflict with a future change to structured logging, rather I think that a log filter is very pythonic to get advanced logging done.

PR #238, opened 2019-06-07
Miscellaneous fixes & refactoring

rhoerbe:
I collected various items that where not related to particular topics separated out into other PRs. 

Note: Commit #fe674be was a mistake; instead of a rebase I rolled it back with a later commit.


rhoerbe:
has been split into other PRs

PR #240, opened 2019-06-14
[microservice] Ldap attribute store, more connection parameters in configuration

peppelinux:
This inherits LdapAttributeStore. 
This is a microservice for binding to a LDAP server that do not accept anonymous connections.

LDAP Connection pool needs anonymous bind for an async and reusable connections.
With this microservice no more slapd ACL modification is required to get stored LDAP attributes from a "hardened" slapd configuration.

Tested on OpenLDAP, Debian10

### All Submissions:


src/satosa/micro_services/ldap_attribute_store_no_pool.py
```diff
@@ -0,0 +1,195 @@
...
+from satosa.exception import SATOSAError
+
+import copy
+import logging
+import ldap3
+import urllib
+
+from ldap3.core.exceptions import LDAPException
+
+from . ldap_attribute_store import *
```
c00kiemon5ter: why do you need this?

src/satosa/micro_services/ldap_attribute_store_no_pool.py
```diff
@@ -0,0 +1,195 @@
...
+
+logger = logging.getLogger(__name__)
+
+class LdapAttributeStoreNoPool(LdapAttributeStore):
```
c00kiemon5ter: Instead of inheriting and creating a new micro-service, can this be a config option in the existing code?

peppelinux: I got issues to handle this kind of approach but we can try some better code to get this result. 
I don't know if these issues could depend by a different version of ldap3, I used ldap3==2.6.

Using the clientstrategy RESTARTABLE I didn't get a valid message_id from connection.search
````
message_id = connection.search(config['search_base'], search_filter, attributes=config['search_return_attributes'].keys())
````
We can do some better code generalization using a more detailed config file if you're interested to this, sure.

src/satosa/micro_services/ldap_attribute_store_no_pool.py
peppelinux: fixed in my last amend, thank you

src/satosa/micro_services/ldap_attribute_store_no_pool.py
peppelinux: Done in https://github.com/IdentityPython/SATOSA/pull/240/commits/121fc799fc76e89b8394b51992548e91bc3721df

Tested in my environment, actually we have two conflicts. The first one is trivial, the second one is more a programming style decision. You'll tell me what direction if you agree.

Beyond these the most important things are here https://github.com/IdentityPython/SATOSA/pull/240/files#diff-6885077c6ac78f71c00e9192de67a5ebR446 
and here https://github.com/IdentityPython/SATOSA/pull/240/files#diff-6885077c6ac78f71c00e9192de67a5ebR468

Regarding these lastest two, I don't know if this behaviour could be linked to ldap3 version used or from the client_strategy method (message_id). So, before merging I have to ask to you a simple test in an environment with `client_strategy=ldap3.REUSABLE`, as that code in the past was made to be working with it. After your feedback we could do the final refactor before merge if you agree this line.

skoranda:
Hello Giuseppe,

I do not understand the basis for this pull request. You state

"LDAP Connection pool needs anonymous bind for an async and reusable connections." I do not believe that is the case. 

The current code uses client_strategy=ldap3.REUSABLE when creating an instance of Connection(), but it also uses a bind DN and a bind password. So the code assumes one is NOT using an anonymous bind.

Perhaps I am not understanding. Can you explain how the current code fails or does not work for you?

Thanks,

Scott

peppelinux:
Hi @skoranda, the usability of REUSABLE strategy depends by slapd ACL. In my case, where anon bind Is disabled, that strategy doesn't work. RESTARTABLE strategy makes It to work and I also suppose that this connection strategy could solve your issue, related to the Need to restart satosa everytime slad restarts, if I understand well.

By the way with this PR we moved ldap3 connection parameters to microservice's configuration, instead of the current code that have many of these as constants in the code.

skoranda:
Hi @peppelinux,

You state "the usability of REUSABLE strategy depends by slapd ACL. In my case, where anon bind Is disabled, that strategy doesn't work."

I am sorry, I do not understand.

Here is the olcAccess attribute for a slapd directory I am using:

olcAccess: {0}to dn.base="cn=admin,dc=myproject,dc=org" by sockname.regex=/
 var/run/slapd/ldapi auth by users none by * none
olcAccess: {1}to attrs=userPassword,shadowLastChange by self auth by anonymous
  auth by * none
olcAccess: {2}to dn.subtree="o=MYPROJECT,dc=myproject,dc=org" by dn="uid=registry
 _user,ou=system,o=MYPROJECT,dc=myproject,dc=org" write by dn="uid=proxy_user,ou=
 system,o=MYPROJECT,dc=myproject,dc=org" read by * none
olcAccess: {3}to * by * none

I am using the existing LDAP attribute store microservice code successfully with that slapd olcAccess configuration.

Can you show me please an olcAccess configuration you have that does NOT work with the current code? I think that will help me understand (and reproduce) the issue you are having.

Thanks,

Scott

peppelinux:
Hi @skoranda, thank you for sharing the configuration, really appreciate.

This is the template of my slapd acl used for tests:
https://github.com/peppelinux/ansible-slapd-eduperson2016/blob/master/roles/slapd_configure/templates/acl.ldif

using client_strategy.REUSABLE I get the following.
Satosa log:
````
un 24 14:44:48 debian10-pysaml-ldap2 satosa-saml2[466]: ldap3.core.exceptions.LDAPResponseTimeoutError: no response from worker threads in Reusable connection
Jun 24 14:44:48 debian10-pysaml-ldap2 satosa-saml2[466]: The above exception was the direct cause of the following exception:
Jun 24 14:44:48 debian10-pysaml-ldap2 satosa-saml2[466]: Traceback (most recent call last):
Jun 24 14:44:48 debian10-pysaml-ldap2 satosa-saml2[466]:   File "/opt/django-saml2-idp.env/lib/python3.7/site-packages/satosa/proxy_server.py", line 123, in __call__
Jun 24 14:44:48 debian10-pysaml-ldap2 satosa-saml2[466]:     resp = self.run(context)
Jun 24 14:44:48 debian10-pysaml-ldap2 satosa-saml2[466]:   File "/opt/django-saml2-idp.env/lib/python3.7/site-packages/satosa/base.py", line 303, in run
Jun 24 14:44:48 debian10-pysaml-ldap2 satosa-saml2[466]:     raise SATOSAUnknownError("Unknown error: {}".format(err)) from err
Jun 24 14:44:48 debian10-pysaml-ldap2 satosa-saml2[466]: satosa.exception.SATOSAUnknownError: Unknown error: no response from worker threads in Reusable connection
````

slapd log
````
[24-06-2019 14:46:11] slapd debug  conn=1026 op=9 SEARCH RESULT tag=101 err=0 nentries=1 text=
[24-06-2019 14:46:11] slapd debug  conn=1026 op=10 BIND anonymous mech=implicit ssf=0
[24-06-2019 14:46:11] slapd debug  conn=1026 op=10 BIND dn="cn=admin,dc=testunical,dc=it" method=128
[24-06-2019 14:46:11] slapd debug  conn=1026 op=10 BIND dn="cn=admin,dc=testunical,dc=it" mech=SIMPLE ssf=0
[24-06-2019 14:46:11] slapd debug  conn=1026 op=10 RESULT tag=97 err=0 text=
[24-06-2019 14:46:11] slapd debug  conn=1026 op=11 UNBIND
[24-06-2019 14:46:11] slapd debug  conn=1026 fd=14 closed
```` 

As we can see the first bind in REUSABLE strategy is an anon bind, then follows the one with a real dn.

skoranda:
@peppelinux Thanks for sharing those details. Let me look in detail and try to reproduce and I will get back to you as soon as possible. Cheers.

peppelinux:
good to hear this @skoranda, really thanks.
The goal of this PR that evolved after ivan's change requests, is the need to expose connection paramenters in ldap store plugin configuration. Even if a conn strategy is needed or not, usable or not, for one or many users, with this PR we can have a customizable ldap client behaviour. 
I need it for the things we discussed, maybe those conn paramenters could be usefull for others users in many different ways.

peppelinux:
Regarding our last f2f meeting, would we like to introduce `is_required` configuration param to implement the possibility to ignore, or not, an ldap connection/query failure?

This could depend about the need, or not, of account linking matches, after the authentication response.

In addition to this, @skoranda talked about the possibility to use context object (the user's session) to share attributes and states of them. I used the same ldap_store plugin, configured in different ways, into the same stack of proxy microservices, to get different behaviour or attribute processing during different state of the same flow. What he purposed could be helpfull in this.

peppelinux:
Hi Scott, this Is why async conn fails on that server

https://github.com/peppelinux/ansible-slapd-eduperson2016/blob/master/roles/slapd_configure/templates/disallow_anonimous_bind.ldif

I also saw that even if REUSABLE Is faster in connection handling the result set come as a decoded dictionary, instead of object as RESTARTABLE does. This latter give us raw datas in a faster way. I found the objects and its methods more flexible and faster during the trailing processing. Shared as it come, bye

PR #249, opened 2019-07-09
 code style: string concatenation

peppelinux:
Simple code style consideration, less is more.


c00kiemon5ter:
This introduces the exact thing that I want to avoid; manually arranging that whitespace at the beginning or ending of each string.

Also, try the formatting that will come out of `black`.

Issue #236, opened 2019-06-03
flagging an ERROR although OK: satosa.state:cookie_to_state:No cookie named SATOSA_STATE


rhoerbe:
SAML AuthRequests do not need the state cookie, actually it dos not make sense to read an existing one.

## Expected Behavior
The endpoint module should know if a state cookie is required and raise an exception if it was missing. satosa.state:cookie_to_state should return an empty state object, not raise an exception

As a result ERRORs would only appear in the log if justified.

peppelinux:
@rhoerbe I completely agree with you, I patched this behaviour in my environment.

![Selezione_579](https://user-images.githubusercontent.com/1297620/59421143-680d6b00-8dce-11e9-9f85-1de8965e7d88.png)

peppelinux:
@rhoerbe sorry for the late, I finally made this PR https://github.com/IdentityPython/SATOSA/pull/250 with my proposal

c00kiemon5ter:
Fixed by fcd61a2d13ad6e7a1dd0e8998ae9490d8ef2ce30 and dcec6e57a30d18f9f9502b9725cb603fff229fc7

PR #250, opened 2019-07-09
A fix for "No cookie named SATOSA_STATE in" Exception

peppelinux:
This PR fixes this frequent error messages in the logs:

````
[2019-07-09 12:53:30] [ERROR]: No cookie named SATOSA_STATE in 
Traceback (most recent call last):
  File "/opt/django-saml2-idp.env/lib/python3.7/site-packages/satosa/state.py", line 74, in cookie_to_state
    state = State(cookie[name].value, encryption_key)
KeyError: 'SATOSA_STATE'
````

Also related to https://github.com/IdentityPython/SATOSA/issues/236

### All Submissions:


peppelinux:
Three test failed, like:

`_ TestStateAsCookie.test_cookie_to_state_handle_broken_cookies[Set-Cookie: state_cookie="_Td6WFoAAATm1rRGAgAhARYAAAB0L-WjAQCXYWt4NU9ZLWF5amdVVDdSUjhWdnkyUHE5MFhJV0J4Uzg5di1EVW1nNTR0WHZKakFsaWJmN2JMOUtlNEltMkJ0dmxOakRyUDJXZE53d0dwSGNqYnBzVng5YjVVeUYyUzkwcWVSMU42U2VNNHZDQTktUXdCQWx0WUh6LVBPX1pBYnZ1M1RsV09Qc2lKS3VpelB5a0FsMG93PT0AmlSCX0Pk2WoAAbABmAEAAGRNyZ2xxGf7AgAAAAAEWVo="; Max-Age=600; Path=/; Secure-state_cookie-wrong_encrypt_key-Exception] _`

because with this fix the unexistent cookies or which do not have the name SATOSA_STATE (or whatever name configured in proxy_conf) will be simply ignored. 

Decide if change the unit tests to handle the null cookie as valuable or remove these unit tests.

c00kiemon5ter:
I opted to remove logging, as it doesn't seem to be of much value. 
Fixed by fcd61a2d13ad6e7a1dd0e8998ae9490d8ef2ce30 

The `cookie_to_state(cookie_str, name, encryption_key)` function is expected to load data from a cookie given a _name_. If that is not found, it should error out, as it already does. Defaulting to another value is not correct; it is unexpected. The call should be notified that what they asked is not available and then they should be the ones to handle what should happen.
In our case the caller is `_load_state` and for that caller not finding the given name is fine, as it can fallback to an empty `State()`. For other callers this is not guaranteed and we should not make such a decision for everyone in `cookie_to_state`.

peppelinux:
Ok, I'd preferred a commit on top of mine Just to track my contributions but you're the boss and everything Is allright. Thank you

PR #252, opened 2019-07-11
Enhance the LDAP attribute store micro-service

skoranda:
A number of commits to enhance the LDAP attribute store. See the commit message for each commit for details.

### All Submissions:

### DIFF
--- a/src/satosa/micro_services/ldap_attribute_store.py
+++ b/src/satosa/micro_services/ldap_attribute_store.py
@@ -36,23 +36,23 @@ class LdapAttributeStore(ResponseMicroService):
     """
 
     config_defaults = {
-        'bind_dn'                       : None,
-        'bind_password'                 : None,
-        'clear_input_attributes'        : False,
-        'ignore'                        : False,
-        'ldap_identifier_attribute'     : None,
-        'ldap_url'                      : None,
-        'on_ldap_search_result_empty'   : None,
-        'ordered_identifier_candidates' : None,
-        'search_base'                   : None,
-        'search_return_attributes'      : None,
-        'user_id_from_attrs'            : [],
-        'read_only'                     : True,
-        'version'                       : 3,
-        'auto_bind'                     : False,
-        'client_strategy'               : ldap3.RESTARTABLE,
-        'pool_size'                     : 10,
-        'pool_keepalive'                : 10,
+        'bind_dn':                       None,
+        'bind_password':                 None,
+        'clear_input_attributes':        False,
+        'ignore':                        False,
+        'ldap_identifier_attribute':     None,
+        'ldap_url':                      None,
+        'on_ldap_search_result_empty':   None,
+        'ordered_identifier_candidates': None,
+        'search_base':                   None,
+        'search_return_attributes':      None,
+        'user_id_from_attrs':            [],
+        'read_only':                     True,
+        'version':                       3,
+        'auto_bind':                     False,
+        'client_strategy':               ldap3.RESTARTABLE,
+        'pool_size':                     10,
+        'pool_keepalive':                10,
         }
 
     def __init__(self, config, *args, **kwargs):
@@ -75,7 +75,7 @@ class LdapAttributeStore(ResponseMicroService):
 
         # Process the default configuration first then any per-SP overrides.
         sp_list = ['default']
-        sp_list.extend([ key for key in config.keys() if key != 'default' ])
+        sp_list.extend([key for key in config.keys() if key != 'default'])
 
         connections = {}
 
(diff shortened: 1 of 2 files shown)


peppelinux:
As reference to RESTARTABLE/REUSABLE client strategy

https://github.com/cannatag/ldap3/issues/221

skoranda:
Hi @peppelinux. I have no problem with making the RESTARTABLE client strategy possible (your PR was already accepted), and I am also fine with making it the default for a later release, but since it would substantially change the behavior for an existing deployment it needs to be a fully documented change. In my opinion it should be considered a breaking change.

In my testing the REUSABLE client strategy is substantially faster and necessary for SATOSA deployments under significant load (tens of SAML flows per second or more).

peppelinux:
Really thanks Scott but I do not need It, I can use both now and the must important thing Is to have parameters instead of constants. Really appreciate Indeed.

I use this New Born client to do massive tests with differents configurations on one or many LDAP servers:
https://github.com/peppelinux/pyMultiLDAP/blob/master/README.md

 I Just went deeper in ldap3 API and I found a very good sharing with you in this thread. I can't ask more!

src/satosa/micro_services/ldap_attribute_store.py
```diff
@@ -443,7 +436,13 @@ def process(self, context, data):
         # and find asserted values to construct the ordered list of values for
         # the LDAP search filters.
         for candidate in config["ordered_identifier_candidates"]:
-            value = self._construct_filter_value(candidate, data)
+            value = self._construct_filter_value(
+                candidate,
+                data.subject_id,
+                data.subject_type,
+                data.auth_info.issuer,
+                data.attriutes,
```
peppelinux: attributes instead of attriutes :upside_down_face:

src/satosa/micro_services/ldap_attribute_store.py
```diff
@@ -369,58 +369,45 @@ def process(self, context, data):
...
+            for candidate in config["ordered_identifier_candidates"]
+            # Consider and find asserted values to construct the ordered list
+            # of values for the LDAP search filters.
+            for filter_value in [
+                self._construct_filter_value(
+                    candidate,
+                    data.subject_id,
+                    data.subject_type,
+                    issuer,
+                    data.attriutes,
```
peppelinux: attributes instead of attriutes

src/satosa/micro_services/ldap_attribute_store.py
skoranda: Your comment overlapped with me finding it and fixing it in my latest commit.

src/satosa/micro_services/ldap_attribute_store.py
skoranda: Your comment overlapped with me finding it and fixing it in my latest commit.

skoranda:
All tests pass locally with Python 3.7. Something transient happened with Python 3.6 and Travis.

src/satosa/micro_services/ldap_attribute_store.py
```diff
@@ -255,184 +265,199 @@ def _ldap_connection_factory(self, config):
...
+            # Deprecated configuration. Will be removed in future.
+            else config["search_return_attributes"]
+        )
+
+        attributes = defaultdict(list)
+
+        for attr, values in ldap_attributes.items():
+            internal_attr = ldap_to_internal_map.get(attr, None)
+            if not internal_attr and ';' in attr:
+                internal_attr = ldap_to_internal_map.get(attr.split(';')[0],
```
c00kiemon5ter: This is strange. 

What is the format that LDAP uses to represent the attributes it returns? 
Does this mean that there are attribute-names that are concatenated by `;`?

I do not think this is the place to handle this. This should be handle in the place where `ldap_attributes` are defined. We shouldn't have to take this into account when traversing the attributes.

skoranda: The logic is necessary to properly handle LDAP attribute options. If you are not aware of LDAP attribute options see RFC 4512 and 4520, though you may find it more efficient to examine the explanation for how LDAP attribute options are leveraged in the voPerson schema at https://github.com/voperson/voperson/blob/master/voPerson.md#voperson-attribute-options.

The Python ldap3 module properly handles LDAP attribute options and the code as written makes it possible for the person deploying and configuring the microservice to map the attribute values to specific SATOSA internal attributes both when the deployer knows precisely which LDAP attribute option is use and when the deployer does not know which attribute options are in use and wants to collect values for all.

c00kiemon5ter: Alright, so this means that attributes could look like:
```
{
    "some_attribute_name": "some_attribute_value",
    "uid;app-wiki": "plee@myvo.org",
    "mobile;internal": "+1 646 555 1212",
}
```

so, the `;<option>` part is optional and we ignore it for this particular processing.

What I suggest is to move this where `ldap_attributes` are defined. So,

```python
        ldap_attributes = {
            name: value
            for name_options, value in record.get("attributes", {}).items()
            for name in [name_options.split(";", 2)[0]]
        }
```

then use the `attr` from `ldap_attributes` without this check.

skoranda: Here is a use case that the code needs to cover. The Python ldap3 module returns

{
    voPersonApplicationUID;app-vidpeptid: [1128993321],
    voPersonApplicationUID;app-vidpeppn: [scottk@messproject.org],
    voPersonAffiliation;role-111: [staff],
    voPersonAffiliation: [member]
}

A deployer may need this mapped to internal attributes like this:

{
  eptid: [1128993321],
  eppn: [scottk@messproject.org],
  affiliation: [staff, member]
}
 
I don't think your suggestion covers that use case. Does it?

rohe: Hold in mind that an attribute description consists of an attribute type and zero or more attribute options. 
I don't think there is an upper limit as to how many options you can have.

skoranda: Right/correct/agreed.

The current logic in the code accommodates that by iterating through the values and attempting to find mappings to internal SATOSA attributes, going from fine-grained mappings to less fine-grained. So if the deployer knows that a particular option is going to exist, she can make sure it gets mapped to a specific internal attribute. If the deployer doesn't know which options might be returned (or none at all), she can have all the values mapped to an internal attribute.

c00kiemon5ter: So, to make this clear (at least for me), the point here is that the mapping to internal attributes could potentially include the options. As such, the match should happen when looking up the attributes - we cannot avoid this.

PR #253, opened 2019-07-11
add code to handle exceptions without traceback

rhoerbe:
Problem: SATOSA log includes traceback for errors that are sufficiently clear from the error message.
Tracebacks are difficult to handle for line-oriented logs and should be eliminated (but this will take time & effort)

This PR introduces SATOSAErrorNoTraceback and bases SATOSAConfigurationError on it.

The satosa_saml_metadata.py script gets a separate handling.


rhoerbe:
this branch did not contain the intended code changes

PR #258, opened 2019-07-13
routing.py: fix minor bug in config check

rhoerbe:
incorrect condition


rhoerbe:
merge issues, withdrawed

PR #260, opened 2019-07-13
fix directory name clash for venv

rhoerbe:
docker/setup.sh: rename /opt/satosa to /opt/venv to disambiguate

### DIFF
--- a/docker/setup.sh
+++ b/docker/setup.sh
@@ -2,7 +2,7 @@
 
 pip3 install --upgrade virtualenv
 
-virtualenv -p python3 /opt/satosa
-/opt/satosa/bin/pip install --upgrade pip setuptools
-/opt/satosa/bin/pip install /src/satosa/
+virtualenv -p python3 /opt/venv
+/opt/venv/bin/pip install --upgrade pip setuptools
+/opt/venv/bin/pip install /src/satosa/
 


docker/setup.sh
```diff
@@ -2,7 +2,7 @@
 
 pip3 install --upgrade virtualenv
 
-virtualenv -p python3 /opt/satosa
-/opt/satosa/bin/pip install --upgrade pip setuptools
-/opt/satosa/bin/pip install /src/satosa/
+virtualenv -p python3 /opt/venv
```
ioparaskev: maybe venv-satosa would be self-explanatory?

rhoerbe:
would be good as well. As most containers are fine with a single venv, /opt/env should be clear, but no objection to make it more clear.

ioparaskev:
did you test this?

rhoerbe:
I have a large dev branch where I test this. Then I copied the specific change pertaining to the venv into a new branch and submitted the PR.
As discussed in the last web call we should submit changes in small pieces. I relied in travis, but I did not check for a test case.

c00kiemon5ter:
What does the virtualenv directory name clash with?

c00kiemon5ter:
If we apply this then all defaults paths in [start.sh](https://github.com/IdentityPython/SATOSA/blob/de922d7/docker/start.sh) will break, along with the [activation of the virtualenv](https://github.com/IdentityPython/SATOSA/blob/de922d7/docker/start.sh#L34-L35).

c00kiemon5ter:
Then, the [IdentityPython/satosa-developer](https://github.com/IdentityPython/satosa-developer) project, that uses the docker image, will fail as the [compose file](https://github.com/IdentityPython/satosa-developer/blob/master/satosa/compose.yml) assumes [those](https://github.com/IdentityPython/satosa-developer/blob/master/satosa/compose.yml#L28-L30) [paths](https://github.com/IdentityPython/satosa-developer/blob/master/satosa/compose.yml#L38).

rhoerbe:
Well, it is a bit of a special feature:-) To make debugging/tracing easier I copy the project to /src/satosa (as a personal convention from other projects.)

I will withdraw this PR - sorry, nobody else will need this.

c00kiemon5ter:
That's the same thing that IdentityPython/satosa-developer does, but since the prefix is different (`/opt` vs `/src`) there is no clash - it will work fine. 

If I understand this correctly, you would prefer the virtualenv directory to have a name that includes `venv` or `virtualenv` or something that makes it _read_ different than the source code directory name. I find a point to that, and maybe we should do it.

If we do this, let's plan for it and do across the relevant repos, and for the other projects too.

rhoerbe:
Yes, it would be good practice, but it is not important.

peppelinux:
venv folder name and app name should have different names, a good practice that it's almost more like a convention I think

PR #261, opened 2019-07-13
Cleanup and enhance docker start.sh

rhoerbe:
- Allow to specify a configuration file for gunicorn
- Allow to specify a chain link for the https certificate


c00kiemon5ter:
Note, you can set the environment variable `GUNICORN_CMD_ARGS` that allow you to pass additional _command line options_ to `gunicorn`. It is not _the same_ as a configuration file, but it allows for customizability.

rhoerbe:
Wrt readability I would prefer the config file

c00kiemon5ter:
merged as #278

PR #263, opened 2019-07-13
add sequence diagrams for SAML-to-SAML operation

rhoerbe:


peppelinux:
Awesome!

peppelinux:
Hi @rhoerbe I would like to ask you how do you produced these diagrams, which tool or method you used

rhoerbe:
The instruction is in the source files: https://www.websequencediagrams.com

peppelinux:
> The instruction is in the source files: https://www.websequencediagrams.com

I saw it but I missed the way you produced the step. Did you made them by hands?
I thought that there was some debugging tool that build up the sequence in txt format first

rhoerbe:
I see. Use the autologging package to generate traces:

from autologging import traced

@traced
class ...

c00kiemon5ter:
`doc/internals/authnrequ_state.png` is missing the corresponding `.src`. Is that on purpose?

rhoerbe:
This was a mistake. It seems that I missed to add the .src to the repo, it is gone. 

I would not like to remove this diagram from the colelction, as it is good to have when you are working with SATOSA_STATE. OTOH, I would prefer to postpone redoing the src until changes are requested.

rhoerbe:
If I understand correctly, `git log --all --full-history -- **/authrequ.src` searches across all branches .. so it is nowhere in the repo.

c00kiemon5ter:
- `authnresp_state` has a note about #234 which has been merged
- `authnrequ_flow` and `authnresp_flow` present the loop (recursion) in a way that looks like some methods are called for each iteration

I guess you already know about those and we can fix them soon.

---

I would like to produce such diagrams and have them be part of the docs for other areas too.
I will try what you did and see if we can find a way to make this part of the CI, so that they can be autoproduced and stay up to date.

rhoerbe:
Automating the process should be possible:
1. Using some preprocessor, add the imports and trace decorators from autologging for each class or function
2. Configure logging so that trace output is generated
3. Run a function test doing request/response in an interesting configuration.
4. Filter the result for keeping only interesting classes and functions
5. Convert the trace to sequence diagram sources

- Filtering might be done by adding trace markup only to relevant places beforehand
- Some permanent markup may be required to add comments or quirks to make the result appealing

Issue #73, opened 2017-03-31
Create Facebook plugin for SATOSA


jkakavas:
Move Facebook oauth2 implementation out of oauth2.py and into its own plugin and repo

c00kiemon5ter:
This should happen, but it will be part of the plugins overhaul. I am closing this, but with a plan to actually do it.

Issue #72, opened 2017-03-31
Create base oauth2 plugin 


jkakavas:
Decouple base oauth2 plugin from facebook and leave this in core satosa

c00kiemon5ter:
We should refactor the oauth2 clients and provide a common base. New clients can inherit from the base and extend it to satisfy their specific needs. 

This work should cover https://github.com/IdentityPython/SATOSA/issues/73

PR #267, opened 2019-07-28
Adds 3 new configuration options for the Facebook backend

skanct:
The Facebook backend does not request any scopes from Facebook. As a
result, the default scope (public_profile) is assumed and the user is
asked to give permission only for that. This has the effect that even
though in the query to the Facebook graph API we might be asking for the
user's e-mail, the graph API will return only the user data that are
part of the public profile.

Furthermore, if a user has given permission for some scopes, then the
the user response is remembered by Facebook and even if the application
changes its configuration and requests for more scopes, Facebook will
not take this into account, unless the 'auth_type' parameter is set to
'rerequest'

This commit introduces three new configuration options:

- 'auth_type' expects list of auth_types. If none is provided, then the
'auth_type' parameter will not be sent to the authorization_endpoint.

- 'scope' expects a list of scopes. If none is provided, then the scope
element will not be added and 'public_profile' will be assumed (as it
happens now)

- 'graph_endpoint' expects the endpoint for the Facebook graph API. Up
to now the value of the graph_endpoint was hardcoded in the actual code
of the backend. In order to retain backwards compatibility if the
configuration option for the 'graph_endpoint' is not set, then it is
assumed to have the value that used to be hardcoded.

### All Submissions:

### DIFF
--- a/src/satosa/backends/oauth.py
+++ b/src/satosa/backends/oauth.py
@@ -186,7 +186,7 @@ class FacebookBackend(_OAuthBackend):
     The default graph endpoint is for backward compatibility with previous versions of the
     Facebook backend in which the graph endpoint was hardcoded in the code.
     """
-    DEFAULT_GRAPH_ENDPOINT = 'https://graph.facebook.com/v2.5/me'
+    DEFAULT_GRAPH_ENDPOINT = "https://graph.facebook.com/v2.5/me"
 
     def __init__(self, outgoing, internal_attributes, config, base_url, name):
         """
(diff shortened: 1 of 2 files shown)


src/satosa/backends/oauth.py
```diff
@@ -178,6 +182,12 @@ class FacebookBackend(_OAuthBackend):
     Backend module for facebook.
     """
 
+    """
+    The default graph endpoint is for backward compatibility with previous versions of the
+    Facebook backend in which the graph endpoint was hardcoded in the code.
+    """
+    DEFAULT_GRAPH_ENDPOINT = "https://graph.facebook.com/v2.5/me"
```
ioparaskev: I think this line here also addresses issues #73 #72 
So maybe these should be closed or at least be reworded ?

PR #270, opened 2019-08-01
Add attributes from the Swiss eduPerson Schema

hannahshort:
Adding attributes used by Switch AAI to the attributemaps


c00kiemon5ter:
Can those attributes come with any of the three name-formats?
- `urn:oasis:names:tc:SAML:2.0:attrname-format:basic`
- `urn:oasis:names:tc:SAML:2.0:attrname-format:uri`
- `urn:mace:shibboleth:1.0:attributeNamespace:uri`

hannahshort:
I'm not 100% sure. There's limited information in the documentation: https://www.switch.ch/aai/docs/AAI_Attr_Specs.pdf 

I'll reach out to SWITCH and ask

c00kiemon5ter:
I am guessing it is following eduPerson, and should come with a `urn:oasis:names:tc:SAML:2.0:attrname-format:uri` or a `urn:mace:shibboleth:1.0:attributeNamespace:uri`

hannahshort:
Hi again, 

I checked with SWITCH and they clarified that it should only be the saml2 value "According to [SDP-IDP18] in 'SAML V2.0 Interoperability Deployment Profile V2.0 (Draft)' [1] the IdP must use as NameFormat only urn:oasis:names:tc:SAML:2.0:attrname-format:uri"

In addition, I'd like to include two missing lines in the saml frontend/backend example files, to make sure that attributes are picked up from the attributemaps.

c00kiemon5ter:
This is now [part of `pysaml2` attributemaps](https://github.com/IdentityPython/pysaml2/commit/247e2f642b37de90a61c4f5ca49d8403840b6ea2).

Issue #228, opened 2019-04-24
[Error Handling] Users read "Unknow Error" if SATOSA session is expired


peppelinux:
If an end user session is expired the user can go from DS to SATOSA to selected IdP anyway, and then do auth, then come back to SATOSA but get "Unknow Error".

Using [this silly patch](https://github.com/IdentityPython/SATOSA/pull/221) I can repr the error to the users, the message would be "Unknown error: 'SATOSA_BASE'".

## Code Version
master branch, current stable

## Expected Behavior
Probably a better handling of expired session could improve user experience when using SATOSA. This could be also a way to introduce a HTML template for 500 error handling.

## Current Behavior
This is the stacktrace regarding the session expiration. As we can read this could be more accurante and eloquent to tell to the sysadmin: "Don't worry, that session is expired...".

````
[2019-04-24 17:26:56] [ERROR]: [urn:uuid:41986301-2e69-4191-8f8f-cd6239316f78] Uncaught exception
Traceback (most recent call last):
  File "/home/user/django-saml2-idp.env/lib/python3.7/site-packages/satosa/base.py", line 286, in run
    resp = self._run_bound_endpoint(context, spec)
  File "/home/user/django-saml2-idp.env/lib/python3.7/site-packages/satosa/base.py", line 228, in _run_bound_endpoint
    return spec(context)
  File "/home/user/django-saml2-idp.env/lib/python3.7/site-packages/satosa/backends/saml2.py", line 246, in authn_response
    return self.auth_callback_func(context, self._translate_response(authn_response, context.state))
  File "/home/user/django-saml2-idp.env/lib/python3.7/site-packages/satosa/base.py", line 185, in _auth_resp_callback_func
    internal_response.requester = context.state[STATE_KEY]["requester"]
  File "/home/user/django-saml2-idp.env/lib/python3.7/site-packages/satosa/state.py", line 230, in __getitem__
    return self._state_dict[key]
KeyError: 'SATOSA_BASE'
[2019-04-24 17:26:56] [ERROR]: Unknown error: 'SATOSA_BASE'
````

## Steps to Reproduce
Start an AuthRequest to SATOSA frontend (use DS to select the IDP if more then one was configured in metadata), then go buy an icecream and take your time to eat it under the spring sun, then go back to put your credentials or give your data consent, then you'll be redirected to SATOSA's backed but... the "Unknow Error" appears.

peppelinux:
[...]

PR #272, opened 2019-08-09
[WiP] Inconsistent context.state exception handling

peppelinux:
An Exception handling on missing SATOSA_BASE in context.state[STATE_KEY]["requester"].
Probably also related to https://github.com/IdentityPython/SATOSA/issues/228

If a http request come from an arbitrary link, like
````
https://disco.example.org/role/idp.ds?entityID=https%3A%2F%2Fproxy.example.org%2FSaml2%2Fmetadata&return=https%3A%2F%2Fexample.org%2FSaml2%2Fdisco
````

the satosa frontend cannot run `SatosaBase._auth_req_callback_func`, so it never inizializes `state[STATE_KEY] = {"requester": internal_request.requester}`.

This PR would introduce a self explanatory error message to user, this permit them to understand the nature of the problem without calling help-desk support.

I admit that this sounds, as it is now, too much hardcoded, I'd like to start a discussion with you to handle in a definitive way this kind of feature.

A good point of discussion could be the possibility to disallow backend's authnRequest if context.state would be inconsistent, I think that this would be a better strategy (also with a human readable message to users that warns on this)

### All Submissions


peppelinux:
Also related to
https://github.com/IdentityPython/SATOSA/issues/228

From that thread there would be some usefull points of discussion I think

PR #273, opened 2019-08-11
Adds a backend for BitBucket

skanct:
This pull request adds an OAuth2 backend for Bitbucket.

Information about the OAuth2 support by Bitbucket can be found here:

https://developer.atlassian.com/bitbucket/api/2/reference/meta/authentication

Information about the Bitbucket can be found here:

https://developer.atlassian.com/bitbucket/api/2/reference/resource/user

### All Submissions:

### DIFF
--- a/example/plugins/backends/bitbucket_backend.yaml.example
+++ b/example/plugins/backends/bitbucket_backend.yaml.example
@@ -1,0 +1,30 @@
+module: satosa.backends.bitbucket.BitBucketBackend
+name: bitbucket
+config:
+  authz_page: bitbucket/auth/callback
+  base_url:  <base_url>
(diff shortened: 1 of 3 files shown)


c00kiemon5ter:
This seems to be the same as the [`FacebookBackend`](https://github.com/IdentityPython/SATOSA/blob/032ebec/src/satosa/backends/oauth.py#L180) but with different `user_information()` handling. Should we make a common base for the two, inherit and differentiate on the `user_information` method?

c00kiemon5ter:
I've merged this, but we should still refactor the code to create a base for Oauth2 clients.
Also, see https://github.com/IdentityPython/SATOSA/issues/72

PR #274, opened 2019-08-31
Fixes support for extra_scopes in the OIDC frontend

skanct:
`_get_approved_attributes` called `scope2claims` with one argument, the scopes the client had requested.

`scope2claims` can receive two arguments, the second being the `extra_scope_dict`. If that is not defined, then `scope2claims` will use only `SCOPE2CLAIMS`, which has only the `openid`,` profile`, `email`, `address`, `phone` and `offine`_access scopes.

This commit changes the call to scope2claims to include also the `extra_scopes` that may have been added in the Provider's configuration.

### All Submissions:

### DIFF
--- a/tests/satosa/frontends/test_openid_connect.py
+++ b/tests/satosa/frontends/test_openid_connect.py
@@ -1,6 +1,7 @@
 """
 Tests for the SAML frontend module src/frontends/saml2.py.
 """
+import copy
 import json
 from base64 import urlsafe_b64encode
 from collections import Counter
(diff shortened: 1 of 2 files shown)


tests/satosa/frontends/test_openid_connect.py
```diff
@@ -199,9 +291,64 @@ def test_provider_configuration_endpoint(self, context, frontend):
...
+            scope in scopes_supported for scope in ["openid", "email", "eduperson"]
+        )
+
+        # FIXME why is this needed?
+        expected_capabilities["claims_supported"] = set(
+            expected_capabilities["claims_supported"]
+        )
+        provider_config_dict["claims_supported"] = set(
+            provider_config_dict["claims_supported"]
+        )
```
c00kiemon5ter: I am not sure why we need to do this, here. I added this as a workaround. Would you like to look into it more?

c00kiemon5ter: Notice, this is not about the order of the dict-fields, but the order of the list elements. Lists are ordered, so this should not happen. This is ofcourse a serialized version of that object, and I'm not sure how the list is generated.

skanct: Actually we do not need to check again provider_config against the expected_capabilities. We can just check the scopes_supported. The problem here is python 3.5 and the random order of the lists. I wonder though how it did not fail randomly also for the other lists in the dict.

c00kiemon5ter: > The problem here is python 3.5 and the random order of the lists.

No, this is not correct - there is no "random order of the lists". What is was not ordered -and I still think one should not rely on that property- is the order of the keys of dictionaries. Lists are and always were ordered.

PR #275, opened 2019-09-11
Improve current logging for SATOSA

sebulibah:
### All Submissions:

### DIFF
--- a/src/satosa/routing.py
+++ b/src/satosa/routing.py
@@ -65,9 +65,9 @@ class ModuleRouter(object):
         else:
             self.micro_services = {}
 
-        logger.debug("Loaded backends with endpoints: %s" % backends)
-        logger.debug("Loaded frontends with endpoints: %s" % frontends)
-        logger.debug("Loaded micro services with endpoints: %s" % micro_services)
+        logger.debug("Loaded backends with endpoints: {}".format(backends))
+        logger.debug("Loaded frontends with endpoints: {}".format(frontends))
+        logger.debug("Loaded micro services with endpoints: {}".format(micro_services))
 
     def backend_routing(self, context):
         """
@@ -79,7 +79,9 @@ class ModuleRouter(object):
         :param context: The request context
         :return: backend
         """
-        logger.debug("Routing to backend: %s " % context.target_backend)        
+        msg =  "Routing to backend: {backend}".format(backend=context.target_backend)
+        logline = "[{id}] {message}".format(id=context.state, message=msg)
+        logger.debug(logline)        
         backend = self.backends[context.target_backend]["instance"]
         context.state[STATE_KEY] = context.target_frontend
         return backend
@@ -96,7 +98,9 @@ class ModuleRouter(object):
         """
 
         target_frontend = context.state[STATE_KEY]
-        logger.debug("Routing to frontend: %s " % target_frontend)        
+        msg = "Routing to frontend: {frontend} ".format (frontend=target_frontend)
+        logline = "[{id}] {message}".format(id=context.state, message=msg)
+        logger.debug(logline)        
         context.target_frontend = target_frontend
         frontend = self.frontends[context.target_frontend]["instance"]
         return frontend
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
I'm taking a random change:

```diff
-        satosa_logging(logger, logging.DEBUG, "Routing to backend: %s " % context.target_backend, context.state)
+        logger.debug("Routing to backend: %s " % context.target_backend)
```

---

I note two things here:

1. To concatenate strings we should prefer `.format()`

```diff
-        satosa_logging(logger, logging.DEBUG, "Routing to backend: %s " % context.target_backend, context.state)
+        msg = "Routing to backend: {backend}".format(backend=context.target_backend)
+        logger.debug(msg)
```

2. The chosen change does not include `context.state` that was previously part of the content that was logged. Specifically, if we look at the `satosa_logging()` function, we see that _state_ is only used to extract the _session id_. So, let's try to keep that:

```diff
-        satosa_logging(logger, logging.DEBUG, "Routing to backend: %s " % context.target_backend, context.state)
+        msg = "Routing to backend: {backend}".format(backend=context.target_backend)
+        logline = "[{id}] {message}".format(id=state.get("SESSION_ID"), message=msg)
+        logger.debug(logline)
```

src/satosa/routing.py
```diff
@@ -80,7 +79,9 @@ def backend_routing(self, context):
         :param context: The request context
         :return: backend
         """
-        satosa_logging(logger, logging.DEBUG, "Routing to backend: %s " % context.target_backend, context.state)
+        msg =  "Routing to backend: {backend}".format(backend=context.target_backend)
+        logline = "[{id}] {message}".format(id=context.state, message=msg)
```
c00kiemon5ter: This should be `context.state.get("SESSION_ID")`

src/satosa/routing.py
```diff
@@ -97,7 +98,9 @@ def frontend_routing(self, context):
         """
 
         target_frontend = context.state[STATE_KEY]
-        satosa_logging(logger, logging.DEBUG, "Routing to frontend: %s " % target_frontend, context.state)
+        msg = "Routing to frontend: {frontend} ".format (frontend=target_frontend)
```
c00kiemon5ter: No need for this extra whitespace at the end, here.

src/satosa/routing.py
```diff
@@ -136,17 +139,20 @@ def endpoint_routing(self, context):
         :return: registered endpoint and bound parameters
         """
         if context.path is None:
-            satosa_logging(logger, logging.DEBUG, "Context did not contain a path!", context.state)
+            logger.debug("Context did not contain a path!")
```
c00kiemon5ter: There is some extra whitespace at the end of the line here.

Some calls are still missing the session id.

c00kiemon5ter:
Thanks @sebulibah 

This is the first merge towards improving the state of logging in the project. For those interested we will start by removing the `satosa_logging` function, replacing it with direct calls to the logger. Then, we will switch to using, probably, the `structlog` library and starting taking advantage of the features of structured logging.

peppelinux:
Good news, I prefer direct call to logging instances too

src/satosa/routing.py
```diff
@@ -136,17 +141,24 @@ def endpoint_routing(self, context):
...
+        logger.debug(logline)
         path_split = context.path.split("/")
         backend = path_split[0]
 
         if backend in self.backends:
             context.target_backend = backend
         else:
-            satosa_logging(logger, logging.DEBUG, "Unknown backend %s" % backend, context.state)
+            msg = "Unknown backend {}".format(backend)
+            logline = "[{id} {message}".format(id=context.state.get("SESSION_ID"), message=msg)
```
c00kiemon5ter: there was a typo here; the `id` should have been followed by `]`

```diff
- "[{id} {message}"
+ "[{id}] {message}"
```

c00kiemon5ter:
@sebulibah see 053df986ae1e45a938d9de97495474c2aef30530 and 4065cfd7ac2c1802977408f5f8b57d83202bcd9e . 

I have extracted the log format and the way we should get the session id, so it should be easier to follow that pattern and avoid such typos in the future ;)

sebulibah:
> Thanks @sebulibah
> 
> This is the first merge towards improving the state of logging in the project. For those interested we will start by removing the `satosa_logging` function, replacing it with direct calls to the logger. Then, we will switch to using, probably, the `structlog` library and start taking advantage of the features of structured logging.

You're welcome

PR #280, opened 2019-09-23
Allow definition of metadata URL for SAML plugins

saibot94:
We at CERN require SAML metadata URLs to be different than the ones provided by parsing the `entityId`. This comes from a need to support an old and new system concurrently, both of them having the same `

This PR extends the config and allows it to be extended with a `custom_metadata_url`.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -151,7 +151,7 @@ Common configuration parameters:
 | `attribute_profile` | string | `saml` | attribute profile to use for mapping attributes from/to response
 | `entityid_endpoint` | bool | `true` | whether `entityid` should be used as a URL that serves the metadata xml document
 | `acr_mapping` | dict | `None` | custom Authentication Context Class Reference
-| `custom_metadata_url` | string | `custom_metadata/endpoint.xml` | custom metadata endpoint, in the case you want to serve your metadata on an endpoint other than the `entityId`s path. Will not get populated if the attribute is missing or empty, or when the `entityid_endpoint` property is not `true`.  |
+| `metadata_endpoint` | string | `my/metadata/endpoint.xml` | metadataendpoint, used for serving the metadata xml at a path of your choice. This config option acts independently from `entityid_endpoint`. |
 
 The metadata could be loaded in multiple ways in the table above it's loaded from a static
 file by using the key "local". It's also possible to load read the metadata from a remote URL.
--- a/doc/README.md
+++ b/doc/README.md
@@ -151,7 +151,7 @@ Common configuration parameters:
 | `attribute_profile` | string | `saml` | attribute profile to use for mapping attributes from/to response
 | `entityid_endpoint` | bool | `true` | whether `entityid` should be used as a URL that serves the metadata xml document
 | `acr_mapping` | dict | `None` | custom Authentication Context Class Reference
-| `metadata_endpoint` | string | `my/metadata/endpoint.xml` | metadataendpoint, used for serving the metadata xml at a path of your choice. This config option acts independently from `entityid_endpoint`. |
+| `metadata_endpoint` | string | `my/metadata/endpoint.xml` | metadata endpoint, used for serving the metadata xml at a path of your choice. This config option acts independently from `entityid_endpoint`. |
 
 The metadata could be loaded in multiple ways in the table above it's loaded from a static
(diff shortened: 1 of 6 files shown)


ioparaskev:
I'm not sure I understand the purpose of this PR. Why not generate the metadata as described [here](https://github.com/IdentityPython/SATOSA/blob/master/doc/README.md#generate-proxy-metadata) and use nginx/apache to serve them under any path you want?

Besides that, seeing the diff and reading the documentation I think that it's not clear and can be confusing what URL will be used in the end to serve the metadata. To be more specific, the documentation says:

> | `entityid_endpoint` | bool | `true` | whether `entityid` should be used as a URL that serves the metadata xml document

**_entityid should be used as a URL that serves the metadata xml document_**

and in your diff it says:

> | `custom_metadata_url` | string | `custom_metadata/endpoint.xml` | custom metadata endpoint, in the case you want to serve your metadata on an endpoint other than the `entityId`s path. Will not get populated if the attribute is missing or empty, or when the `entityid_endpoint` property is not `true`.  |

_**serve your metadata on an endpoint other than the `entityId`s path**_

So you set the `entityid_endpoint` to mark that **_entityid should be used as a URL that serves the metadata xml document_** and you set `custom_metadata_url` to some other URL _**to serve your metadata on an endpoint other than the `entityId`s path**_. So the second option overrides the first option which is confusing I think.

saibot94:
Thanks for the comments @ioparaskev. We just wanted to have something simple to configure, without necessarily requiring nginx / apache in front of it. 

Anyway, I understand your concern on the confusing aspect of the code. I'll change the implementation so that the `custom_metadata_url` is not dependent on the `entityid_endpoint`. I might also change the name to `custom_metadata_endpoint` to make it more consistent.

c00kiemon5ter:
@saibot94 call it `metadata_endpoint` and we may make this the primary configuration option to server metadata.

saibot94:
I renamed the endpoint and change the description in the docs. Let me know if it makes sense, thanks!

doc/README.md
```diff
@@ -151,6 +151,7 @@ Common configuration parameters:
 | `attribute_profile` | string | `saml` | attribute profile to use for mapping attributes from/to response
 | `entityid_endpoint` | bool | `true` | whether `entityid` should be used as a URL that serves the metadata xml document
 | `acr_mapping` | dict | `None` | custom Authentication Context Class Reference
+| `metadata_endpoint` | string | `my/metadata/endpoint.xml` | metadata endpoint, used for serving the metadata xml at a path of your choice. This config option acts independently from `entityid_endpoint`. |
```
ioparaskev: maybe add "acts independently **and overrides** `entityid_endpoint` option" to be more clear?

c00kiemon5ter: no need, we will probably do the opposite; I'll now explain in the PR discussion

c00kiemon5ter:
So, we now have to options that seem similar; they both handle the URL that will serve the metadata of the corresponding plugin. These are: `entityid_endpoint` (old) and `metadata_endpoint` (new).

`entityid_endpoint` came by the requirement of eIDAS to serve metadata at the location indicated by the entityid. This means that eIDAS specifies that the entityid MUST be a URL.

While this is good practice it is not specified by the SAML specification. The SAML specification says that the entityid is a URI. Any resource identified can be used, and in practice this has been used and deployed with `urn:oid` and other types of ids.

For those entities there is no way to specify a URL that will serve the metadata. The metadata have to be generated separately and served by other means. This is doable and we have been working like this was for a long time.

The new option `metadata_endpoint` covers exactly this need. It solves the problem of serving the metadata for an entity that defines an entityid that is not a URL. The use case of CERN seems different, but if one considers the entityid as just an id, then it is exactly the same problem. The fact they define the entityid as a URL doesn't make the use case any less valid. URLs are subsets of URIs. They define a URI and they need a URL to serve the metadata.

At the same time `metadata_endpoint` is more general than `entityid_endpoint`. This means that it can cover the use case of `entityid_endpoint` and thus it can replace it. So, I propose that we go forward with that. We introduce a new option, and then we'll deprecate the old one.

saibot94:
Hi @c00kiemon5ter! I wanted to ask if there's any status update on this pull and #279 . We'd require this feature available relatively soon. :)

PR #282, opened 2019-10-01
Fixed orcid authentication and access to API issues

sgomez:
Hi, my name is Sergio Gómez and I belongs to the Trust and Identity Incubator from GEANT. We are implementing the ORCID backend and we discover two issues:

1. ORCID supports [bearer token in _Authorization_ header](https://github.com/ORCID/ORCID-Source/blob/master/orcid-api-web/tutorial/read_public.md) to access to the API. Right now, SATOSA is sending two headers (_Authorization type_ and _Access Token_). This is still working in the public API but not in sandbox. _Authorization_ header works in both.

2. Inside address, only country attribute is returned and not _address_ ([example](https://github.com/ORCID/ORCID-Source/blob/master/orcid-core/src/test/resources/test-addresses-2.1.xml)). When a country is sent from the API the module crash.

Linter only fails in some line length.

### All Submissions:

### DIFF
--- a/src/satosa/backends/orcid.py
+++ b/src/satosa/backends/orcid.py
@@ -96,8 +96,7 @@ class OrcidBackend(_OAuthBackend):
         url = urljoin(base_url, '{}/person'.format(orcid))
         headers = {
             'Accept': 'application/orcid+json',
-            'Authorization type': 'Bearer',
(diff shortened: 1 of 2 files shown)


c00kiemon5ter:
thanks!

PR #284, opened 2019-10-07
Add state parameter to ORCID authorization request

sgomez:
OAuth2 state parameter is supported by [ORCID authorization endpoint](https://members.orcid.org/api/resources/customize#state).

This PR modifies ORCID backend to support the state parameter from base class `OAuth2Backend`:

- Removed `start_auth` method to use it from base class.
- Added `get_request_args` method to compose the authorization uri inside base `start_auth` method.
- Refactorized `_authn_response` method. Now is very close to base method.  The only difference is that `auth_info` requires two more attributes from the authorization response (_orcid_ and _name_). But base method only sends _access_token_. A  `user_info` base method refactorization could resolve this in a way than `_authn_response` does not need to be overloaded.
- ORCID test was modified too to check state parameter.

It is tested against ORCID sandbox environment too.

### All Submissions:

### DIFF
--- a/src/satosa/backends/orcid.py
+++ b/src/satosa/backends/orcid.py
@@ -13,7 +13,7 @@ from oic.oauth2.message import AuthorizationResponse
 from satosa.backends.oauth import _OAuthBackend
 from satosa.internal import InternalData
 from satosa.internal import AuthenticationInformation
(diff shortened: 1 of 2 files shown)


c00kiemon5ter:
That's great, thanks!

PR #288, opened 2019-10-17
(WIP) enable microservice-based flow to replay AuthnRequest (requires wsgi_app in registered callback)

rhoerbe:
The rationale for this change is a microservice that is replaying the authentication request to the IDP after receiving the response with a signaling attribute that requires some interaction (like profile completion or role selection. ADFS is not capable of implementing this in the login sequence.

The microservice stores the authnRequest. Once it receives the response from the external interaction, it returns wsgi_app.run(authnrequ_context) to replay the AuthnRequest.


rhoerbe:
After wading thru various plugins that might potentially be called in _run_bound_endpoint() I decided to add **kwargs to all methods with the (self, context [, ...]) signature, because it is not trivial to proof what could be called (except a few, suche as process())

The first commit contains of the PR contains the microservice redirecturl.py, for which this PR is used. It is deleted in the second PR to limit this PR to a single purpose.

PR #300, opened 2019-11-05
Properly change the package version

c00kiemon5ter:
On version `4.5.0` the package version indicator was not changed, but a corresponding branch was created. The version is now changed directly to `v4.5.1` to accommodate for the fix commit.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from setuptools import setup, find_packages
 
 setup(
     name='SATOSA',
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
We only merge this to keep the content consistent. The actual tag `v4.5.0` and thus the packaged version of the code remains under the _branch_ `v4.5.0`'s timeline.

PR #304, opened 2019-11-08
Improve logging - satosa.micro_services.ldap_attribute_store

sebulibah:
### All Submissions:

### DIFF
--- a/src/satosa/micro_services/ldap_attribute_store.py
+++ b/src/satosa/micro_services/ldap_attribute_store.py
@@ -15,11 +15,10 @@ from ldap3.core.exceptions import LDAPException
 from collections import defaultdict
 
 from satosa.exception import SATOSAError
-from satosa.logging_util import satosa_logging
 from satosa.micro_services.base import ResponseMicroService
 from satosa.response import Redirect
 
-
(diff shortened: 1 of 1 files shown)


src/satosa/micro_services/ldap_attribute_store.py
```diff
@@ -66,15 +65,17 @@ def __init__(self, config, *args, **kwargs):
 
         if "default" in config and "" in config:
             msg = """Use either 'default' or "" in config but not both"""
-            satosa_logging(logger, logging.ERROR, msg, None)
+            logline = lu.LOG_FMT.format(id=lu.get_session_id(None), message=msg)
```
c00kiemon5ter: Why do we do `lu.get_session_id(None)`?
I think in all other cases, when `state` was `None` we skipped setting the `id` and just logged the `message`. Is that correct?

Shouldn't this be
```diff
-            satosa_logging(logger, logging.ERROR, msg, None)
+            logger.error(msg)
```

sebulibah: Yes, that's correct. Let me change that.

PR #313, opened 2019-11-26
Better handling of single-value attribute by LdapAttributeStore

skoranda:
Better handling of single-value attributes from LDAP by the
LdapAttributeStore microservice. The ldap3 module response object
includes a raw_attributes dictionary and an attributes dictionary. The
attributes dictionary is populated with formatted values using
formatters based on standard schema syntaxes. If a particular attribute
in the schema is defined as multi valued, the attribute value in the
attributes dictionary is a list, but if the attribute in the schema is
not multi valued, the value in the attributes dictionary is a single
value.

### All Submissions:

### DIFF
--- a/src/satosa/micro_services/ldap_attribute_store.py
+++ b/src/satosa/micro_services/ldap_attribute_store.py
@@ -12,8 +12,6 @@ import urllib
 import ldap3
 from ldap3.core.exceptions import LDAPException
 
-from collections import defaultdict
-
 from satosa.exception import SATOSAError
 from satosa.logging_util import satosa_logging
 from satosa.micro_services.base import ResponseMicroService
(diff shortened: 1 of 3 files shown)


src/satosa/micro_services/ldap_attribute_store.py
```diff
@@ -265,7 +263,22 @@ def _ldap_connection_factory(self, config):
...
+        client_strategy_string = config["client_strategy"]
+        client_strategy_map = {
+            "SYNC": ldap3.SYNC,
+            "ASYNC": ldap3.ASYNC,
+            "LDIF": ldap3.LDIF,
+            "RESTARTABLE": ldap3.RESTARTABLE,
+            "REUSABLE": ldap3.REUSABLE,
+            "MOCK_SYNC": ldap3.MOCK_SYNC
+        }
+        client_strategy = client_strategy_map[client_strategy_string]
```
c00kiemon5ter: this could also be 

```python
client_strategy = getattr(ldap3, config.get("client_strategy", ""), None)
if client_strategy is None:
    raise LdapAttributeStoreError("client_strategy is required")
```

PR #316, opened 2020-02-13
Pull YAML configuration values from environment

skoranda:
Add logic so that a YAML tag of the form !ENV indicates
that a value of the form ${SOME_ENVIRONMENT_VARIABLE} should
be replaced with the value of the process environment variable
of the same name.

### All Submissions:


c00kiemon5ter:
The YAML specification is very specific on the YAML format. YAML does not have extensions (to my knowledge) and cannot read from the environment.

By adding special syntax that is treated differently that the rest of the YAML document, we are effectively creating our own unspecified format. This is wrong and bad.
We are locking our configuration files to the project, as the only way to reliably reuse those files is to enforce this unofficial extension to the other project that will need to look into those files. We cannot use the standard tools reliably, because they will have to understand this new syntax.
We are also locking ourselves to this specific parser, that allows us to do this. PyYAML has had a history of security incidents and is considered slow. We will be replacing it at some point.
At the same time, the environment can only be used to set string-type values. There is no way to set any other complex value.
Practically (and hopefully), this will only be used for "secrets". But, the tradeoffs described, do not seem very beneficial to me.

Instead, if one needs to read values from the environment, one should use a format that actually supports reading values from the environment. Currently, we support YAML files. For some things we also support Python (`.py`) files. YAML cannot read from the environment, but Python can. If you convert your file to a Python file, then you can use the `os` module to read from the environment (`os.environ.get("ENV_VAR_NAME", "FALLBACK")`). 
The point here is that Python does have support for this operation. We may have to add support for loading `.py` files, but this seems way better than doing our own mini-yaml format. If we really need this ability, I think that's where we should focus.

Having said this, there is a big discussion that can be had as to how secure it is or is not to have secrets in files or the environment. What are the tradeoffs and the assumptions for evaluation.

skoranda:
YAML does have the capability of being extended. That is precisely what the "!" is for and how the Python YAML module is capable (out of the box) of consuming Python class instances as configuration.

I would like a larger discussion before this PR is closed.

c00kiemon5ter:
> YAML does have the capability of being extended. That is precisely what the "!" is for 

That's good! I can see this in the [specification](https://yaml.org/spec/1.2/spec.html); "Tags" in YAML-lang. Basing something on this is reasonable. I'll read more on it. Thanks for the pointer.

src/satosa/yaml.py
```diff
@@ -0,0 +1,58 @@
...
+    filepath = os.environ.get(raw_value)
+    if filepath is None:
+        msg = "Cannot construct value from {node}: {path}".format(
+            node=node, path=filepath
+        )
+        raise YAMLError(msg)
+
+    try:
+        with open(filepath, "r") as fd:
+            new_value = fd.read()
```
c00kiemon5ter: `.strip()` doesn't take place anymore.

src/satosa/yaml.py
```diff
@@ -0,0 +1,58 @@
...
+    else:
+        return new_value
+
+
+TAG_ENV = "!ENV"
+TAG_ENV_FILE = "!ENVFILE"
+
+
+_safe_loader.add_constructor(TAG_ENV, _constructor_env_variables)
+_safe_loader.add_constructor(TAG_ENV_FILE, _constructor_envfile_variables)
```
c00kiemon5ter: We only have value constructors; explicitly. We do not resolve anything implicitly.

doc/README.md
```diff
@@ -29,9 +29,23 @@ apt-get install libffi-dev libssl-dev xmlsec1
 Alternatively the application can be installed directly from PyPI (`pip install satosa`), or the [Docker image](https://hub.docker.com/r/satosa/) can be used.
 
 # Configuration
+SATOSA is configured using YAML.
+
 All default configuration files, as well as an example WSGI application for the proxy, can be found
 in the [example directory](../example).
 
+A configuration value that includes the tag !ENV will have a value of the form `SOME_ENVIRONMENT_VARIABLE`
```
c00kiemon5ter: No more `${}` or `$()`. We only define the env-var name.

c00kiemon5ter:
@skoranda see the changes I made.

I think if we add documentation for the `!ENVFILE` tag, this should be ready.

src/satosa/yaml.py
c00kiemon5ter: Can we assume that everyone will want the contents of a file to be `.strip()`ed?

_Not_ doing this, means that the user can later choose to do so, later; it also means that the user is responsible for the file contents, and that content should be properly catered if no such whitespace is wanted.
Doing this here, means that the user has no chance to restore the stripped characters.

skoranda:
By removing the use of ${} and $() to indicate the text being replaced you have removed functionality like

```
some_config: !ENV /do/not/replace/this/${BUT_DO_REPLACE_THIS}

c00kiemon5ter:
hm, I will think this more, but

> `/do/not/replace/this/${BUT_DO_REPLACE_THIS}`

is effectively mixing ideas from different places (re: shell expansion). I would prefer to keep this clean, ie `!ENV` defines a variable that will be read from the environment. I do not think it is hard for anyone to set the variable to the whole like, ie set the env to `FOO="/do/not/replace/this/something"` and then have the yaml part `!ENV FOO` to resolve to the complete string.

skoranda:
How about going with the code as is (thanks for your improvements and refactoring) and later if we find a direct use case for "shell expansion" then we can reconsider?

I am more interested in getting this first round committed and released than worrying over the details.

c00kiemon5ter:
I want to think about the mixed-content/shell-expansion a bit more to decide.
I am not opposed to it, but I would prefer to base it on actual use-cases.

Right now, it would be weird to have `!ENV` define a string, part of which will be replaced by an env-var content, but have `!ENVFILE` not do the same.

I think it is best to keep things simple, and do more complex replacements as needed.
If we really need to do in-place replacements, we can later define a new tag to signify that specific purpose.

PR #317, opened 2020-02-21
Specify pool name when using REUSABLE client strategy

skoranda:
Add code to choose a random name to be used as the pool name when using
the REUSABLE client strategy. Without this enhancement two instances of
the microservice connecting to distinct servers will share a single
connection pool.

### All Submissions:

No new tests because the pool name is only used for real connections to real LDAP servers, not during mock ups.

### DIFF
--- a/src/satosa/micro_services/ldap_attribute_store.py
+++ b/src/satosa/micro_services/ldap_attribute_store.py
@@ -7,6 +7,8 @@ the record and assert them to the receiving SP.
 
 import copy
 import logging
+import random
+import string
 import urllib
 
 import ldap3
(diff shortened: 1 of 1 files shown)


src/satosa/micro_services/ldap_attribute_store.py
```diff
@@ -303,6 +305,8 @@ def _ldap_connection_factory(self, config):
 
         pool_size = config["pool_size"]
         pool_keepalive = config["pool_keepalive"]
+        pool_name = ''.join(random.sample(string.ascii_lowercase, 6))
```
c00kiemon5ter: I don't remember the entropy that `random` has; another idea may be to use `uuid.uuid4()`.

c00kiemon5ter:
Is this ready for merge?

src/satosa/micro_services/ldap_attribute_store.py
skoranda: There are no crytographic considerations here. The random number is only used to ensure that two pools have different names. So the only requirement is that success calls to random() deliver N distinct numbers where N is the number of LDAP servers being contacted. It seems likely that N will be < 10 and random() is easily able to meet the requirement.

skoranda:
Yes, ready for merge. Thanks.

PR #324, opened 2020-04-16
Context state keyError Exception message for humans

peppelinux:
This PR introduces a human readable message to Users when they gets SATOSA_BASE KeyError.
Found it usefull to avoid weird Users open assistane tickets when they gets that "ermetic" error on the screen.

### DIFF
--- a/src/satosa/base.py
+++ b/src/satosa/base.py
@@ -134,6 +134,12 @@ class SATOSABase(object):
         """
 
         context.request = None
+        context_state = context.state.get(STATE_KEY)
+        if not context_state:
+            redirect_url = self.config.get("UNKNOW_ERROR_REDIRECT_PAGE")
+            raise SATOSAStateError(('context.state has no {}. Your session '
+                                    'is not valid, please start a new '
+                                    'Authentication request again.'.format(STATE_KEY)))
         internal_response.requester = context.state[STATE_KEY]["requester"]
 
         # If configured construct the user id from attribute values.


ioparaskev:
While I'm in favor of a more helpful error message in the logs (or in a FAQ maybe), the current log entry is like it is addressed to the user and not the admin that will look at the error.
In addition to that I think that it's better to handle error pages/redirects outside SATOSA to have less concerns and let the admin customize how they want to handle such cases

p.s: in the current commit while there's a `redirect_url`, it's not used anywhere

peppelinux:
Thank you John, you are proof of the existence of life here :')

What we would do to get this job done?
My users now have something clear, a hint and we would also move to a better design, so I agree with you.

Would we instead use a redirect page?
We should specialize another url for this case or reuse a more general one, like I did here:
https://github.com/IdentityPython/SATOSA/pull/325

space for ideas, the integrations are often trivial

src/satosa/base.py
```diff
@@ -134,6 +134,12 @@ def _auth_resp_callback_func(self, context, internal_response):
         """
 
         context.request = None
+        context_state = context.state.get(STATE_KEY)
+        if not context_state:
```
c00kiemon5ter, line 138: Which are the conditions that make this possible? Why would the state be empty?

src/satosa/base.py
```diff
@@ -134,6 +134,12 @@ def _auth_resp_callback_func(self, context, internal_response):
...
+            redirect_url = self.config.get("UNKNOW_ERROR_REDIRECT_PAGE")
```
c00kiemon5ter, line 139: This needs to be removed as @ioparaskev mentioned.

src/satosa/base.py
```diff
@@ -134,6 +134,12 @@ def _auth_resp_callback_func(self, context, internal_response):
...
+            raise SATOSAStateError(('context.state has no {}. Your session '
+                                    'is not valid, please start a new '
+                                    'Authentication request again.'.format(STATE_KEY)))
```
c00kiemon5ter, line 142: I think it is good to catch such cases or exceptions, but first we should make sure we understand the error, where it comes from and why it is happening. Can we change the code to make it as if the error did not happen at all?

c00kiemon5ter: As @ioparaskev said, SATOSA is currently not concerned about user-facing messages. This message is a mix of technical ("context.state has no") and user-facing information ("please start a new authentication"). I think that neither part is good. What we want to provide in an exception is not a message, but error context. Which information is useful to understand why this happened? This is the information that the exception should hold. I lean on making this information structured, a dict with meaningful keys and data as values.

src/satosa/base.py
peppelinux: This is a superior knowledge that I would gain when I grow up ;)

That's the footprint I collected:
https://github.com/IdentityPython/SATOSA/pull/326#issuecomment-620572473

Which method do you suggest to do, should we pick all the things with a python debugger or we should move to a big view that let us to satisfy our users without make too much noise in the code. It's a kind of paternalism that we acquire getting older, the users are tring to tell us something ... 

Probably you know better  than me how and when the context.state became inconsistent. Let's try to talk about this is a separate thread if you agree

src/satosa/base.py
peppelinux: @ioparaskev @c00kiemon5ter I miss something here, should we really print the things only on the logs and do not redirect users to a informational page? 

We could find a transitional solution if you agree, something good for our users and then a following refactor with more design elements in the hands. Let me know, you know my limits, try your best :)

src/satosa/base.py
peppelinux: I think that the PR suggested by @skoranda is the solution for the 60% of this kind of problem, really. I follow the users in their daily abits and many of them use the Disco url to bookmark the SP... That's it.

Scott gave us the remediation here:
https://github.com/IdentityPython/SATOSA/pull/322

Move in that direction, merge that first then come back here with better refactoring. I'm behind you (but with 2 meters of covid-social-distancing!)

peppelinux: ping here

c00kiemon5ter:
closed by d9864643a5c606b2e2da14c0f5d6c5271178b5ac

peppelinux:
See my comment here
https://github.com/IdentityPython/SATOSA/pull/363#issuecomment-1586322365

PR #325, opened 2020-04-19
HTTP Redirect page on UnknowError Exception

peppelinux:
This PR introduces a new proxy_config.yml parameter called UNKNOW_ERROR_REDIRECT_PAGE, a possibility to redirect to that page all the Users that gets UnknowError Exception in the Http Response.

- Added a UnknownSystemEntity message for humans in satosa.base (We should also use a settings_messages.py file and import them all from there if you agree)
- Added exceptions.SATOSAUnknownErrorRedirectUrl

### DIFF
--- a/src/satosa/base.py
+++ b/src/satosa/base.py
@@ -10,7 +10,7 @@ from saml2.s_utils import UnknownSystemEntity
 from satosa import util
 from .context import Context
 from .exception import SATOSAConfigurationError
-from .exception import SATOSAError, SATOSAAuthenticationError, SATOSAUnknownError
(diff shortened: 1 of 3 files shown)


peppelinux:
this is another place where a redirect page could be useful

https://github.com/IdentityPython/SATOSA/blob/38dd8a3f9ea5ee8f3e207dcfe910e443d595fb22/src/satosa/proxy_server.py#L98

c00kiemon5ter:
closed by d9864643a5c606b2e2da14c0f5d6c5271178b5ac

peppelinux:
See my comment here
https://github.com/IdentityPython/SATOSA/pull/363#issuecomment-1586322365

PR #328, opened 2020-05-22
AuthnContextClassRef in context.state

oernst9:
* adding AuthnContextClassRef to the context.state in order to make it available to microservices (specifically for the WebAuthn microservice)

* The reason why I needed this change is that the microservice has to decide whether to trigger the MFA or not

### All Submissions:

### DIFF
--- a/src/satosa/frontends/saml2.py
+++ b/src/satosa/frontends/saml2.py
@@ -189,6 +189,10 @@ class SAMLFrontend(FrontendModule, SAMLBaseModule):
         req_info = idp.parse_authn_request(context.request["SAMLRequest"], binding_in)
         authn_req = req_info.message
         msg = "{}".format(authn_req)
+        if "AuthnContextClassRef" in msg:
+            authn_context = msg[msg.find("<ns1:AuthnContextClassRef>") +
+                                26:msg.find("</ns1:AuthnContextClassRef>")]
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
At this point in the code, the value of the `AuthnContextClassRef` can be found at 

```
authn_req.requested_authn_context.authn_context_class_ref[0].text
```

However, you should note that the parent `RequestedAuthnContext` element can hold multiple `AuthnContextClassRef` elements (thus in the above line, we select the first/0th element.) So, the correct approach here, is to create a list from those values:

```
authn_context = [
    ref.text
    for ref in authn_req.requested_authn_context.authn_context_class_ref
]
```

This is how you should be extracting that information. This is the first step.

Your actual goal is to make this available in the micro-services. To do this you should make it part of the context. The context is a `satosa.context.Context` object, that provides the `decorate` and `get_decoration` methods. We typically, define a key (`KEY_..`) that will be used for the new information we store and then we reference the key name. This helps as we can know which keys are expected to be used.

The same way, later on we store `Context.KEY_METADATA_STORE`

```
...
context.decorate(Context.KEY_METADATA_STORE, self.idp.metadata)
...
```

Would you like to change the code towards that direction?

c00kiemon5ter:
fixed by 2801eb1a19feddf5e8571a383368cfc30b683295

PR #332, opened 2020-06-10
Add dynamic requested attributes

nsklikas:
This MR introduces the feature of populating the requested attributes of the SAML backend's request based on the attributes that were requested from the frontend.

### All Submissions:

### DIFF
--- a/tests/satosa/backends/test_saml2.py
+++ b/tests/satosa/backends/test_saml2.py
@@ -211,10 +211,14 @@ class TestSAMLBackend:
 
             kwargs = mock.call_args[1]
             assert "requested_attributes" in kwargs
-            assert kwargs["requested_attributes"] == [
-                {"friendly_name": "cn", "required": True},
-                {"friendly_name": "sn", "required": False}
-            ]
+            assert all(
+                any(r == req for r in requested_attributes)
+                for req in kwargs["requested_attributes"]
+            )
+            assert (
+                len(kwargs["requested_attributes"])
+                == len(requested_attributes)
+            )
 
     def test_authn_request_requested_attributes_ignore_extra(
         self, context, idp_conf, sp_conf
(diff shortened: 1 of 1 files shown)


nsklikas:
This PR has several problems:
- As @c00kiemon5ter pointed this uses the eidas RequestedAttributes extention and not the SAML2 extention (http://docs.oasis-open.org/security/saml-protoc-req-attr-req/v1.0/saml-protoc-req-attr-req-v1.0.html). So perhaps it should be moved to the eidas backend? The saml backend already has the option to use `requested_attributes` (statically configured) which also uses the eIDAS extention.
- In the case of eIDAS this is not correct. The eidas message format states:
```
For attributes requested and being mandatory
according to the eIDAS minimum data sets and [eIDAS-Attr-Profile] the attribute isRequired of
<eidas:RequestedAttribute> MUST be set to “true”. For all optional attributes according
to the eIDAS minimum data sets and [eIDAS-Attr-Profile] the attribute isRequired of<eidas:RequestedAttribute> MUST be set to “false”
```
So for each attribute we need to know not only the name but also whether it is required as well. We could require the operator to provide a list of attributes, in the format of the existing requested attributes, that will be if they were requested from the frontend. This behaviour will be useful in the case of eIDAS as we can't always request all the attributes with one call as sometimes we might want the natural person attributes and other times the legal person attributes.
- If an internal attibute maps to more than one saml attributes the first one will be required. But I'm not sure this is actually a problem.

nsklikas:
I have fixed some of the aforementioned problems, here are a couple of notes on the changes:
- The operator will have to provide the attributes he wants to be requested, i.e.:
```
  dynamic_requested_attributes:
    - friendly_name: attr1
      required: True
    - friendly_name: attr2
      required: False
```
This is needed because eidas requires that we use the correct value for required (i.e. we can't have `is_required=False` for an attribute which is required and vice versa). The only other option would be to add a required field in the attribute mapping (?), but this doesn't sound very clean. Should required be optional?
- Perhaps the `_get_requested_attributes` should be moved to the eIDAS backend. The SAML backend's function should be an empty function, until (and if) the saml extension is implemented in pysaml2.

src/satosa/backends/saml2.py
```diff
@@ -232,13 +243,59 @@ def construct_requested_authn_context(self, entity_id):
...
-    def authn_request(self, context, entity_id):
+    def _get_requested_attributes(self, requested_attributes):
+        if not requested_attributes:
+            return
+
+        attrs = self.converter.from_internal_filter(
+            self.attribute_profile, requested_attributes
+        )
+        requested_attrs = []
+        for attr in attrs:
```
ioparaskev, line 254: I think this can be also written like the following to avoid the nested loops and the appending to list:
```python
from itertools import product

attrs_req_attrs_product = product(attrs, self.requested_attributes)

requested_attrs = [
    dict(friendly_name=attr, required=req_attr['required'])
    for (attr, req_attr) in attrs_req_attrs_product 
    if req_attr['friendly_name'] == attr
]
```

PR #335, opened 2020-07-25
Added RegexSubProcessor Attribute Processor

senorkrabs:
I created a new attribute processor that can substitute and reformat attribute values using using regex patterns. This PR adds a new AttributeProcessor class, and also includes an sample configuration.

### All Submissions:


c00kiemon5ter:
Ideally, in the future, the core of satosa will be separated from the micro-services/plugins. The plugins will be their own package and can be provided by anyone (you can already do this). I am accepting this, but keep in mind that at some point, a new breaking release will come out which will be only about the core - details are not yet clear, but I'm giving you a direction of where we'll go moving forward.

PR #338, opened 2020-09-29
Add sub_hash_salt to README and example configuration

melanger:
Missing documentation.
Note that `sub_hash_salt` is regenerated on startup if not specified in config, which results in varying identifiers.
https://github.com/IdentityPython/SATOSA/blob/8b641cebbc4910ecc5ac897a67e1e530cf408c24/src/satosa/frontends/openid_connect.py#L99
Not needed.
Not needed.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -433,6 +433,7 @@ The configuration parameters available:
 * `signing_key_path`: path to a RSA Private Key file (PKCS#1). MUST be configured.
 * `db_uri`: connection URI to MongoDB instance where the data will be persisted, if it's not specified all data will only
    be stored in-memory (not suitable for production use).
+* `sub_hash_salt`: salt which is hashed into the `sub` claim. If it's not specified, SATOSA will generate a random salt on each startup, which means that users will get new `sub` value after every restart.
 * `provider`: provider configuration information. MUST be configured, the following configuration are supported:
     * `response_types_supported` (default: `[id_token]`): list of all supported response types, see [Section 3 of OIDC Core](http://openid.net/specs/openid-connect-core-1_0.html#Authentication).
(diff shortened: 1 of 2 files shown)


c00kiemon5ter:
This should be a micro-service, not part of the core. There is no reason to enforce such a policy.

SATOSA is used not only to communicate with entities externally, but also within organizations, where the clear values are needed. In such an environment SATOSA acts mostly as a central policy enforcer on all supported protocols.

A micro-service for this should be very easy, and in fact we already provide a micro-service that works around the same idea. 
see, [the hasher micro-service](https://github.com/IdentityPython/SATOSA/blob/master/src/satosa/micro_services/hasher.py)

Note that the hasher micro-service does not generate a new salt, rather one has to be provided. If you feel that generating a salt is needed I would consider a PR for this. I would however, make the behaviour configurable.

Thanks for all your comments so far!

melanger:
> This should be a micro-service, not part of the core. There is no reason to enforce such a policy.

SATOSA hashes the sub claim in default configuration without any microservice, it cannot be turned off.
If you configure no microservices and no salt, it still hashes the sub claim, and it uses a random salt.
If you configure the salt to be an empty string, it still hashes the sub claim.
This is an undocumented behavior, so it would make sense to me to either stop SATOSA doing that (as you said, provide the clear values) or include my patch for the README.
You can inspect the code of openid_connect.py, the class `HashBasedSubjectIdentifierFactory` is used regardless of configuration.

c00kiemon5ter:
You are right, this is undocumented. It is the behaviour of the pyop-based OIDC frontend. Sorry, I misunderstood the initial intention.

If you want to provide a `SubjectIdentifierFactory` that is not the `HashBasedSubjectIdentifierFactory`, and make it configurable, feel free to do that and I will merge the PR. Whether a value should be hashed should be a choice, not a default behaviour.

Keep in mind that a new OIDC frontend, based on oidcendpoint, is being developed that behaves better, and supports more operations. The OIDC (pyop-based) frontend will continue to exist side by side to the new one, but focus is given there. This should be coming in the following two months.

Thanks!

PR #349, opened 2021-02-01
Make the ScopeExtractorProcessor usable for the Primary Identifier

### LINKED ISSUE #348: Unable to use `ScopeExtractorProcessor` with the primary identifier
## Code Version
7.0.1, but I saw the problem also in the current master.

## Expected Behavior
I would expect that the `ScopeExtractorProcessor` (and other processors) can properly handle the primary identifier so that from, e.g. `myname@mydomain.org`, the domain `mydomain.org` is properly extracted.

## Current Behavior
If extracting the scope from the primary identifier, the processor thinks the input is a `list` instead of a `str` and iterates over the chars and tries to extract the scope, which results in the scope being

```python
>>> '@'.split('@')[1]
''
```

## Possible Solution
You can distinct in the processor between a `str` and a `list` like in the `ScopeProcessor`.

## Steps to Reproduce

Use the `ScopeExtractorProcessor` on the primary identifier.

shaardie:
This patch adds support to use the ScopeExtractorProcessor on the Primary
Identifiert which is, in contrast to the other values, a string.

Closes #348

### All Submissions:

### DIFF
--- a/src/satosa/micro_services/processors/scope_extractor_processor.py
+++ b/src/satosa/micro_services/processors/scope_extractor_processor.py
@@ -31,6 +31,8 @@ class ScopeExtractorProcessor(BaseProcessor):
         values = attributes.get(attribute, [])
         if not values:
             raise AttributeProcessorWarning("Cannot apply scope_extractor to {}, it has no values".format(attribute))
+        if not isinstance(values, list):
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
thanks!

PR #352, opened 2021-02-02
New Microservice `AttributePolicy`

shaardie:
This patch introduces a new micro_service, which is able to force attribute
policies for requester by limiting results to a predefined set of allowed
attributes.

Sometimes you want to make sure that only specific attributes can be received by a specific requester.

### All Submissions:

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -566,6 +566,18 @@ the string `"foo:bar"`:
         "attr1": "foo:bar"
 ```
 
+#### Apply a Attribute Policy
+
+Attributes delivered from the target provider can be filtered based on a list of allowed attributes per requester
+using the `AttributePolicy` class:
+```yaml
+attribute_policy:
+    <requester>:
+        allowed:
+            - attr1
+            - attr2
+```
+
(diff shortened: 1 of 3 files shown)


doc/README.md
```diff
@@ -566,6 +566,18 @@ the string `"foo:bar"`:
         "attr1": "foo:bar"
 ```
 
+#### Apply a Attribute Policy
+
+Attributes delivered from the target provider can be filtered based on a list of allowed attributes per requester
+using the `AttributePolicy` class:
+```yaml
+attribute_policy:
+    <requester>:
```
c00kiemon5ter, line 575: Some notes:

- very often you may want to configure a micro-service per-requester (as you do here), but also per issuer or even per registration-authority
- this usually takes the form of a dict within a dict

```
<scoped-option>:
  <issuer-or-registration-authority-or-default>:
    <requester-or-default>:
      <config-option>: <value(s)>
```

- pysaml2 has the ability to do this using the policy configuration, but I can understand how doing this through a micro-service is easier and more flexible and can grow to support more complex conditions

PR #353, opened 2021-02-02
Make the ID token lifetime configurable

shaardie:
This patch adds a new configuration option to the pyop Provider making it
possible to configure the lifetime of the ID token.

Sometime you need to configure the lifetime of the ID token. E.g. we are using the `JWT` for our web proxy and we need them to be valid as long as our session last.

### All Submissions:

* [x] Does your submission pass tests? **I did not test the mongodb stuff**

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -442,7 +442,7 @@ The configuration parameters available:
     * `access_token_lifetime`: how long access tokens should be valid, see [default](https://github.com/SUNET/pyop#token-lifetimes)
     * `refresh_token_lifetime`: how long refresh tokens should be valid, if not specified no refresh tokens will be issued (which is [default](https://github.com/SUNET/pyop#token-lifetimes))
     * `refresh_token_threshold`: how long before expiration refresh tokens should be refreshed, if not specified refresh tokens will never be refreshed (which is [default](https://github.com/SUNET/pyop#token-lifetimes))
-    * `id_token_lifetime`: the lifetime of the ID token in seconds, see [default](https://github.com/SUNET/pyop#token-lifetimes)
+    * `id_token_lifetime`: the lifetime of the ID token in seconds - the default is set to 1hr (3600 seconds) (see [default](https://github.com/SUNET/pyop#token-lifetimes))
 
 The other parameters should be left with their default values.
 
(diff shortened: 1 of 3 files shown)


doc/README.md
```diff
@@ -442,6 +442,7 @@ The configuration parameters available:
     * `access_token_lifetime`: how long access tokens should be valid, see [default](https://github.com/SUNET/pyop#token-lifetimes)
     * `refresh_token_lifetime`: how long refresh tokens should be valid, if not specified no refresh tokens will be issued (which is [default](https://github.com/SUNET/pyop#token-lifetimes))
     * `refresh_token_threshold`: how long before expiration refresh tokens should be refreshed, if not specified refresh tokens will never be refreshed (which is [default](https://github.com/SUNET/pyop#token-lifetimes))
+    * `id_token_lifetime`: the lifetime of the ID token in seconds, see [default](https://github.com/SUNET/pyop#token-lifetimes)
```
c00kiemon5ter, line 445:
```suggestion
    * `id_token_lifetime`: the lifetime of the ID token in seconds - the default is set to 1hr (3600 seconds) (see [default](https://github.com/SUNET/pyop#token-lifetimes))
```

example/plugins/frontends/openid_connect_frontend.yaml.example
```diff
@@ -12,5 +12,6 @@ config:
     scopes_supported: ["openid", "email"]
     extra_scopes:
       foo_scope:
-	  - bar_claim
-	  - baz_claim
+      - bar_claim
+      - baz_claim
+    id_token_lifetime: 600
```
c00kiemon5ter, line 17:
```suggestion
    id_token_lifetime: 3600
```

src/satosa/frontends/openid_connect.py
```diff
@@ -93,6 +93,7 @@ def _create_provider(self, endpoint_baseurl):
             cdb,
             Userinfo(self.user_db),
             extra_scopes=extra_scopes,
+            id_token_lifetime=self.config["provider"].get("id_token_lifetime", 600),
```
c00kiemon5ter, line 96:
```suggestion
            id_token_lifetime=self.config["provider"].get("id_token_lifetime", 3600),
```

c00kiemon5ter (changes requested):
pyop sets the default ID token lifetime to 1 hour. I think we can keep this as the default. 
600secs/10mins seems low.

This would be a breaking change if we set something different before, but since this is was unset, the defaults were considered. We preserve this functionality but now, we expose the option to be configured by the users.

shaardie:
I agree, I just misread and thought the default were 600.

c00kiemon5ter (approved):

PR #354, opened 2021-02-02
Separate client db configuration from user-data db configuration

shaardie:
This patch adds a new option `prefer_client_db_file` to provide the possibility
to prefer the file over the MongoDB for the client database.

We are deploying Satosa in Kubernetes and it's easier to provide the client database via a file. It enables us to do GitOps and have the configuration handled by git.

The new configuration defaults to `false`, so if not explicitly set, nothing changes here.

### All Submissions:

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -430,6 +430,8 @@ The configuration parameters available:
 * `signing_key_path`: path to a RSA Private Key file (PKCS#1). MUST be configured.
 * `db_uri`: connection URI to MongoDB instance where the data will be persisted, if it's not specified all data will only
    be stored in-memory (not suitable for production use).
+* `client_db_uri`: connection URI to MongoDB instance where the client data will be persistent, if it's not specified the clients list will be received from the `client_db_path`.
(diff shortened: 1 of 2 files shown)


c00kiemon5ter:
I may be missing something but, if you don't set `db_uri` then `client_db_path` will be loaded.

Are you trying to use both the db and the file as a seed?

shaardie:
I want to use the file for the clients. But aren't there several other things stored in the database? So I want that database, too.

c00kiemon5ter:
OK, what you want is to configure the clients-database differently than the "user-data"-database. The database contains authorization codes, access tokens, refresh tokens, subject identifiers, etc.
I think we should make that separation explicit:

- remove `prefer_client_db_file` flag
- additionally to `client_db_path`, introduce `client_db_uri`
- initialize `cdb` by looking at `client_db_uri` and `client_db_path` or in-memory

If you don't set `client_db_uri` then `client_db_path` will be used (assuming it is set) and you'll have the desired result.
If `client_db_uri` is set and `db_uri` is set, then we have the same effect as if setting `db_uri` before.

This will be a breaking change, but easy to overcome - just set the `db_uri` and `client_db_uri` to the same value.

Is this aligned to what you want?
Do you want to make those changes?

shaardie:
Sounds good. I will implement it that way.

shaardie:
@c00kiemon5ter I changed the code according to your description.

PR #358, opened 2021-03-12
Add flake8 Config in Tox

shaardie:
This patch adds a config section for flake8 to tox and also add an ignore for
long lines, since it seems like these errors are already actively ignored.

Some IDEs (tested only vscode) will pick this up and make working a lot easier
and clearer.

### All Submissions:


c00kiemon5ter:
I think that's good to have.
What I'd like to start using is `black`, so that we don't have to worry about formatting, line length, quotes, etc.

PR #361, opened 2021-03-15
Set Primary Identifier only if it exists

shaardie:
This patch ensures that the primary identifier is only set, if it is present
and not `None`.

### All Submissions:

### DIFF
--- a/src/satosa/micro_services/primary_identifier.py
+++ b/src/satosa/micro_services/primary_identifier.py
@@ -250,13 +250,14 @@ class PrimaryIdentifier(satosa.micro_services.base.ResponseMicroService):
             logger.debug(logline)
             data.attributes = {}
 
-        # Set the primary identifier attribute to the value found.
-        data.attributes[primary_identifier] = primary_identifier_val
-        msg = "{} Setting attribute {} to value {}".format(
-            logprefix, primary_identifier, primary_identifier_val
-        )
-        logline = lu.LOG_FMT.format(id=lu.get_session_id(context.state), message=msg)
-        logger.debug(logline)
+        if primary_identifier:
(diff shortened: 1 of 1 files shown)


src/satosa/micro_services/primary_identifier.py
```diff
@@ -250,13 +250,14 @@ def process(self, context, data):
...
             data.attributes = {}
 
-        # Set the primary identifier attribute to the value found.
-        data.attributes[primary_identifier] = primary_identifier_val
-        msg = "{} Setting attribute {} to value {}".format(
-            logprefix, primary_identifier, primary_identifier_val
-        )
-        logline = lu.LOG_FMT.format(id=lu.get_session_id(context.state), message=msg)
-        logger.debug(logline)
+        if primary_identifier:
```
c00kiemon5ter, line 253: From what I understand (I've never used this micro-service), `primary_identifier` will be falsy, only when `config['primary_identifier']` or `self.config['primary_identifier']` are set to a falsy value. If that happens then the assignment may become problematic.

Should this ever happen? Is it OK to have `config['primary_identifier']` set to a falsy value?
Or should we make sure this never happens, which would make this check unnecessary?

shaardie: I am not an expert either, but we am using Satosa with SAML2 as backend with several different IDPs. Some of those are not under our control. The IDPs can decide on their own which attributes they return. We are trying different attributes as primary identifier, but it could happen that none of those are set. So I ended up with something like `"userid": None` in the attributes, which does not make sense.

I took a look at the code and did not find any other place were the primary identifier was used, so I thought it was better to simply not set it. What else could be done for this situation?

PR #365, opened 2021-04-11
feat: added some useful http headers to context

peppelinux:
This PR adds in satosa context object some useful HTTP request headers in a private attribute called `_http_headers`.
these are the following

- REQUEST_METHOD
- PATH_INFO
- REQUEST_URI
- QUERY_STRING
- SERVER_NAME
- REMOTE_ADDR
- HTTP_HOST
- HTTP_USER_AGENT
- HTTP_ACCEPT_LANGUAGE

With these additional values we would allow some highly specialized feature in the future Microservices.
Some example:

- Security: SIEM integrations, checks upon SERVER_NAME, QUERY_STRING, HTTP_USER_AGENT
- i18n: HTTP_ACCEPT_LANGUAGE

### All Submissions:

### DIFF
--- a/src/satosa/proxy_server.py
+++ b/src/satosa/proxy_server.py
@@ -65,6 +65,13 @@ def unpack_request(environ, content_length=0):
     return data
 
 
+def unpack_http_headers(environ):
+    headers = ('REQUEST_METHOD', 'PATH_INFO', 'REQUEST_URI',
+               'QUERY_STRING', 'SERVER_NAME', 'REMOTE_ADDR',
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
@peppelinux we will be merging this PR.

I think the extracted headers could also be configurable. We will re-iterate to do this and probably handle it outside of the proxy-application but as its own middleware.

peppelinux:
> @peppelinux we will be merging this PR.
> 
> I think the extracted headers could also be configurable. We will re-iterate to do this and probably handle it outside of the proxy-application but as its own middleware.

seems quite resonable to me, thank you

PR #366, opened 2021-04-11
feat: IdPHinting Microservice - minimal idp_hinting support

peppelinux:
This PR allows to circumvent the discovery service screen if a SP that supports idp_hinting includes the idp_hint parameter, or alternative name, in the HTTP parameters of its request. One or more names of this parameter can be defined within the configuration of this new Micro Service.

### Example
The following URL triggers an AuthnRequest to SATOSA, with the idp_hinting parameter

http://sp.example.org/saml2/login/?idp=https://172.17.0.1:10000/Saml2IDP/metadata&next=/saml2/echo_attributes&idphint=http%253A%252F%252Flocalhost%253A8080

The SP that supports idp_hinting used for tests is [djangosaml2](https://djangosaml2.readthedocs.io/).

### Without idp hinting
satosa1

### With IdP Hinting
satosa2

### Dependency
This PR depends on this PR:
https://github.com/IdentityPython/SATOSA/pull/365

### All Submissions:

### DIFF
--- a/src/satosa/micro_services/idp_hinting.py
+++ b/src/satosa/micro_services/idp_hinting.py
@@ -1,0 +1,59 @@
+import logging
+from urllib.parse import parse_qs
+
+from .base import RequestMicroService
+from ..exception import SATOSAConfigurationError
+from ..exception import SATOSAError
+
+
+logger = logging.getLogger(__name__)
+
+
+class IdpHintingError(SATOSAError):
+    """
+    SATOSA exception raised by IdpHinting microservice
+    """
+    pass
+
+
+class IdpHinting(RequestMicroService):
+    """
+    Detect if an idp hinting feature have been requested
+    """
+
+    def __init__(self, config, *args, **kwargs):
+        """
+        Constructor.
+        :param config: microservice configuration
+        :type config: Dict[str, Dict[str, str]]
+        """
+        super().__init__(*args, **kwargs)
+        try:
+            self.idp_hint_param_names = config['allowed_params']
+        except KeyError:
+            raise SATOSAConfigurationError(
+                f"{self.__class__.__name__} can't find allowed_params"
+            )
+
+    def process(self, context, data):
+        """
+        This intercepts if idp_hint paramenter is in use
+        :param context: request context
+        :param data: the internal request
+        """
+        target_entity_id = context.get_decoration(context.KEY_TARGET_ENTITYID)
+        qs_raw = context._http_headers['QUERY_STRING']
+        if target_entity_id or not qs_raw:
+            return super().process(context, data)
+
+        qs = parse_qs(qs_raw)
+        hints = (
+            entity_id
+            for param in self.idp_hint_param_names
+            for entity_id in qs.get(param, [None])
+            if entity_id
+        )
+        hint = next(hints, None)
+
+        context.decorate(context.KEY_TARGET_ENTITYID, hint)
+        return super().process(context, data)
--- a/example/plugins/microservices/idp_hinting.yaml.example
+++ b/example/plugins/microservices/idp_hinting.yaml.example
@@ -1,0 +1,6 @@
+module: satosa.micro_services.idp_hinting.IdpHinting
+name: IdpHinting
+config:
+  allowed_params:
+    - idp_hinting
+    - idp_hint


src/satosa/micro_services/idp_hinting.py
```diff
@@ -0,0 +1,57 @@
...
+        :param data: the internal request
+        """
+        if not all((not context.internal_data.get('target_entity_id'),
+                    context._http_headers.get('QUERY_STRING'))):
+            return super().process(context, data)
+
+        qs = urllib.parse.parse_qs(context._http_headers['QUERY_STRING'])
+        for pn in self.idp_hint_param_names:
+            if pn in qs and qs.get(pn):
+                # exit on first match
```
c00kiemon5ter, line 51: The idea is that `idp_hint_param_names` is a list of query parameter names ordered by precedence/preference.

peppelinux: It's correct and these one are defined here:

https://github.com/IdentityPython/SATOSA/pull/366/files#diff-a6f38640ecfe5c5f3b435a2c8990356a443ac5e4a96045ee493555799eda2dd4R4

src/satosa/micro_services/idp_hinting.py
```diff
@@ -0,0 +1,57 @@
...
+                entity_id = qs[pn][0]
+                context.internal_data['target_entity_id'] = entity_id
+                context.request['entityID'] = entity_id
+                break
+
+        return super().process(context, data)
```
c00kiemon5ter, line 57: - we gather the hints
- we select the first one
- we use it to set the target-idp (`target_entity_id`)

If we do not have a proper hint, target_entity_id is set to `None`, which should behave the same as being unset.

```suggestion
        hints = (
            entity_id
            for param in self.idp_hint_param_names
            for entity_id in qs.get(param, [None])
            if entity_id
        )
        hint = next(hints, None)

        context.decorate(Context.KEY_TARGET_ENTITYID, hint)
        return super().process(context, data)
```

src/satosa/micro_services/idp_hinting.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+        This intercepts if idp_hint paramenter is in use
+        :param context: request context
+        :param data: the internal request
+        """
+        if not all((not context.internal_data.get('target_entity_id'),
+                    context._http_headers.get('QUERY_STRING'))):
+            return super().process(context, data)
+
+        qs = urllib.parse.parse_qs(context._http_headers['QUERY_STRING'])
```
c00kiemon5ter, line 48:
```suggestion
        qs = urllib.parse.parse_qs(qs_raw)
```

src/satosa/micro_services/idp_hinting.py
```diff
@@ -0,0 +1,58 @@
+import logging
+import urllib
```
c00kiemon5ter, line 2:
```suggestion
from urllib.parse import parse_qs
```

c00kiemon5ter:
> This PR depends on this PR: #365

#365 has been merged ;)

src/satosa/micro_services/idp_hinting.py
```diff
@@ -0,0 +1,58 @@
...
+            )
+
+    def process(self, context, data):
+        """
+        This intercepts if idp_hint paramenter is in use
+        :param context: request context
+        :param data: the internal request
+        """
+        if not all((not context.internal_data.get('target_entity_id'),
+                    context._http_headers.get('QUERY_STRING'))):
```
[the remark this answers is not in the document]
peppelinux: Done: https://github.com/peppelinux/SATOSA/commit/870e9385512d3e4fc10e0cad8b2088fc7d776dea

src/satosa/micro_services/idp_hinting.py
peppelinux: Done: https://github.com/peppelinux/SATOSA/commit/870e9385512d3e4fc10e0cad8b2088fc7d776dea

src/satosa/micro_services/idp_hinting.py
peppelinux: Done: https://github.com/peppelinux/SATOSA/commit/870e9385512d3e4fc10e0cad8b2088fc7d776dea

peppelinux:
@c00kiemon5ter thank you for your revision, I made my tests, this PR looks ok

peppelinux:
@c00kiemon5ter something goes wrong with the last commits, using hinting with decidebackendbyidp ms.
I'm doing my checks right now

peppelinux:
this is needed 
https://github.com/IdentityPython/SATOSA/pull/366/commits/000f1b38b5b89a8dfc71a2c0a0e3a25f52b98593

c00kiemon5ter:
> this is needed
> [000f1b3](https://github.com/IdentityPython/SATOSA/commit/000f1b38b5b89a8dfc71a2c0a0e3a25f52b98593)

Why is it needed though?

c00kiemon5ter:
The only place where I see this used in on `disco_response` to extract the answer from the discovery service

https://github.com/IdentityPython/SATOSA/blob/9874cb2/src/satosa/backends/saml2.py#L369-L373

How is this related to with this PR?

peppelinux:
> > this is needed
> > [000f1b3](https://github.com/IdentityPython/SATOSA/commit/000f1b38b5b89a8dfc71a2c0a0e3a25f52b98593)
> 
> Why is it needed though?

Otherwise decidebackendbytargetidp won't intercept the request

c00kiemon5ter:
This PR is focused to IdP-hinting. I do not see how `decidebackendbytargetidp` is related. That is a different micro-service that should do what it needs to do on its own.

With IdP-hinting, what we say is that we can read a hint from the query-params and set the target-entity-id to that, so that we do not send the user to the discovery service, but directly to the hinted-IdP.
The purpose of IdP-hinting is to bypass the discovery step.

c00kiemon5ter:
see also, https://github.com/IdentityPython/SATOSA/pull/220/files#r668705488

---

> that's why I'd configure the request entityID to the hinted idp

The `request['entityID']` shouldn't be used anywhere, but the discovery-response handler to extract the selected IdP from the discovery page. Once we set the target_entity_id we shouldn't be caring what the discovery does, because we should be bypassing it.

On DecideBackendByTargetIdP, I see that you get the entityID from that place, but this is an assumption that doesn't hold. The selected IdP is defined by the target_entity_id (`context.get_decoration(Context.KEY_TARGET_ENTITYID)`) as set by this (IdpHinting) or other micro-services (before the DecideBackendByTargetIdP micro-service was invoked).

peppelinux:
got it, thank you

PR #367, opened 2021-05-14
[draft] OIDC frontend: support Redis and session expiration

bajnokk:
This PR adds support for all storage backends in pyop (see https://github.com/IdentityPython/pyop/pull/32). Add automatic expiration TTL for the different collections, so that the session databases do not grow without bounds. The SATOSA lifetime defaults have been copied from pyop defaults, however, advice is needed for those collections for which pyop does not set a default lifetime.

The documentation was just 'monkeypatched', mentioning Redis as an alternative in `db_uri`.

TODO: update pyop version requirement once there is an official release.

### All Submissions:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
     packages=find_packages('src/'),
     package_dir={'': 'src'},
     install_requires=[
-        "pyop >= 3.0.1",
+        "pyop >= 3.0.1", # TODO
         "pysaml2 >= 6.5.1",
         "pycryptodomex",
         "requests",
--- a/doc/README.md
+++ b/doc/README.md
@@ -423,14 +423,14 @@ Connect Relying Parties (RPs). The default configuration file can be found
 [here](../example/plugins/frontends/openid_connect_frontend.yaml.example).
 
 As opposed to the other plugins, this plugin is NOT stateless (due to the nature of OpenID Connect using any other
-flow than "Implicit Flow"). However, the frontend supports using a MongoDB instance as its backend storage, so as long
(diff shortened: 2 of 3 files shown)


peppelinux:
@bajnokk I like the generalization of the storage engine you introduce in this PR!

bajnokk:
Followed up changes in https://github.com/IdentityPython/pyop/pull/32.

c00kiemon5ter:
> #XXX What is the correct ttl for user_db? Is it the same as authz_code_db?
> ...
> ttl=self.config["provider"].get("authorization_code_lifetime", 600),
> ...
> ttl=self.config["provider"].get("access_token_lifetime", 3600),
> ...
> ttl=self.config["provider"].get("refresh_token_lifetime", None),
> ...
> #XXX what is the correct TTL for sub_db?

@bajnokk we can set them to `None` but let users configure those to whatever they like.

setup.py
```diff
@@ -15,7 +15,7 @@
     packages=find_packages('src/'),
     package_dir={'': 'src'},
     install_requires=[
-        "pyop >= 3.0.1",
+        "pyop >= 3.0.1", # TODO
```
c00kiemon5ter, line 18:
```suggestion
        "pyop >= 3.2.0",
```

pyop now needs redis or mongo from extra requires. We can bubble up this choice here, too:

```
    extras_require={
        "ldap": ["ldap3"],
        "redis": ["pyop[redis]"],
        "mongo": ["pyop[mongo]"],
    },
```

We should now be able to do `pip install satosa[redis]` and get the correct dependencies.

_(needs to be tested)_

c00kiemon5ter: This is fixed by #375

PR #369, opened 2021-06-03
Expose metadata endpoint and fix duplicate entity ids with multiple backends for VirtualCoFrontend

vkadam:
### All Submissions:

While using VirtualCoFrontend module, if exposed `entityid_endpoint` and have multiple backend names then metadata exposed is wrong as entityIds generated for frontend are not unique.

This PR fixes
- allow exposing `entityid_endpoint` for VirtualCoFrontend
- ability to make entity id unique which having multiple backends
- backward compatibility for existing name pattern of entityId when using VirtualCoFrontend with single backend.


c00kiemon5ter:
> ability to make entity id unique which having multiple backends

I haven't looked into this yet, but from your description, in a way, this sounds a lot like the `SAMLMirrorFrontend` but for the `VirtualCoFrontend`. Is this correct?

vkadam:
That's correct, unique id was implemented for VirtualFrontend, but it had issues and was not producing unique ids when used with multiple backends and multiple frontends. This PR fixes that issue.

PR #370, opened 2021-06-11
SAML2 frontend+backend: support reloading metadata

vladimir-mencl-eresearch:
Using the reload_metadata method added into pysaml2 in IdentityPython/pysaml2#809, support reloading metadata when triggered via an externally exposed URL (as `/<module_name>/reload-metadata`)

This is off by default (URL not exposed) and needs to be explicitly enabled by setting the newly introduced config option `enable_metadata_reload` for the SAML modules to `true` (or `yes`).

The loaded config is already preserved in the modules, so can be easily used to provide a reference copy of the metadata configuration to the `reload_metadata` method.

This is implemented separately for the SAML2 Backend and SAML2 Frontend (applying to all three SAML2 Frontend classes).

This will complete the missing functionality identified in IdentityPython/pysaml2#808

### All Submissions:

* [Not yet] Have you written new tests for your changes?
* [N/A] Does your submission pass tests?

### DIFF
--- a/src/satosa/backends/saml2.py
+++ b/src/satosa/backends/saml2.py
@@ -82,6 +82,7 @@ class SAMLBackend(BackendModule, SAMLBaseModule):
     KEY_SAML_DISCOVERY_SERVICE_URL = 'saml_discovery_service_url'
     KEY_SAML_DISCOVERY_SERVICE_POLICY = 'saml_discovery_service_policy'
     KEY_SP_CONFIG = 'sp_config'
+    KEY_METADATA = 'metadata'
     KEY_SEND_REQUESTER_ID = 'send_requester_id'
     KEY_MIRROR_FORCE_AUTHN = 'mirror_force_authn'
     KEY_MEMORIZE_IDP = 'memorize_idp'
@@ -479,7 +480,21 @@ class SAMLBackend(BackendModule, SAMLBaseModule):
             url_map.append(("^{0}".format(parsed_entity_id.path[1:]),
                             self._metadata_endpoint))
 
+        if self.enable_metadata_reload():
+            url_map.append(
+                ("^%s/%s$" % (self.name, "reload-metadata"), self._reload_metadata))
+
         return url_map
+
+    def _reload_metadata(self, context):
+        """
+        Reload SAML metadata
+        """
+        logger.debug("Reloading metadata")
+        res = self.sp.reload_metadata(copy.deepcopy(self.config[SAMLBackend.KEY_SP_CONFIG][SAMLBackend.KEY_METADATA]))
+        message = "Metadata reload %s" % ("OK" if res else "failed")
+        status = "200 OK" if res else "500 FAILED"
+        return Response(message=message, status=status)
 
     def get_metadata_desc(self):
         """
--- a/src/satosa/base.py
+++ b/src/satosa/base.py
@@ -261,6 +261,7 @@ class SATOSABase(object):
 
 class SAMLBaseModule(object):
     KEY_ENTITYID_ENDPOINT = 'entityid_endpoint'
+    KEY_ENABLE_METADATA_RELOAD = 'enable_metadata_reload'
     KEY_ATTRIBUTE_PROFILE = 'attribute_profile'
     KEY_ACR_MAPPING = 'acr_mapping'
     VALUE_ATTRIBUTE_PROFILE_DEFAULT = 'saml'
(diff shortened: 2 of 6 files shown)


src/satosa/base.py
```diff
@@ -276,6 +277,15 @@ def expose_entityid_endpoint(self):
...
 
+    def enable_metadata_reload(self):
+        """
+        Check whether metadata reload has been enabled in config
+
+        return: bool
+        """
+        value = self.config.get(self.KEY_ENABLE_METADATA_RELOAD, False)
+        return bool(value)
+
```
c00kiemon5ter, line 288: Add generic form

```suggestion

    def _reload_metadata(self, context, config):
        """
        Reload SAML metadata
        """
        logger.debug("Reloading metadata")
        res = self.idp.reload_metadata(copy.deepcopy(config))
        message = "Metadata reload %s" % ("OK" if res else "failed")
        status = "200 OK" if res else "500 FAILED"
        return Response(message=message, status=status)
```

src/satosa/backends/saml2.py
```diff
@@ -479,8 +480,22 @@ def register_endpoints(self):
...
 
+    def _reload_metadata(self, context):
+        """
+        Reload SAML metadata
+        """
+        logger.debug("Reloading metadata")
+        res = self.sp.reload_metadata(copy.deepcopy(self.config[SAMLBackend.KEY_SP_CONFIG][SAMLBackend.KEY_METADATA]))
+        message = "Metadata reload %s" % ("OK" if res else "failed")
+        status = "200 OK" if res else "500 FAILED"
+        return Response(message=message, status=status)
```
c00kiemon5ter, line 497: use generic form

```suggestion
        config = self.config[SAMLBackend.KEY_SP_CONFIG][SAMLBackend.KEY_METADATA]
        return super()._reload_metadata(context, config)
```

src/satosa/frontends/saml2.py
```diff
@@ -484,6 +491,16 @@ def _metadata_endpoint(self, context):
...
 
+    def _reload_metadata(self, context):
+        """
+        Reload SAML metadata
+        """
+        logger.debug("Reloading metadata")
+        res = self.idp.reload_metadata(copy.deepcopy(self.config[SAMLFrontend.KEY_IDP_CONFIG][SAMLFrontend.KEY_METADATA]))
+        message = "Metadata reload %s" % ("OK" if res else "failed")
+        status = "200 OK" if res else "500 FAILED"
+        return Response(message=message, status=status)
```
c00kiemon5ter, line 502: use generic form

```suggestion
        config = self.config[SAMLFrontend.KEY_IDP_CONFIG][SAMLFrontend.KEY_METADATA]
        return super()._reload_metadata(context, config)
```

src/satosa/base.py
c00kiemon5ter: The generic form has a different signature

```
    def _reload_metadata(self, context, config):
```

vs

```
    def _reload_metadata(self, context):
```

maybe `SAMLBackend._reload_metadata` and `SAMLFrontend._reload_metadata` should have a different name.

On the suggestions, I call `super()._reload_metadata` to differentiate the two. If they are renamed, we can call `self._reload_metadata(context, config)` directly.

c00kiemon5ter (changes requested):
I've made some suggestions on the structure. Let me know what you think.

vladimir-mencl-eresearch:
Hi @c00kiemon5ter ,

Thanks also for this review.

I agree with the motivation for a generic method - I myself wasn't so comfortable with duplicating the code between the two copies of `_reload_metadata` in `SAMLBackend` and `SAMLFrontend`.

The two methods have very few differences - but besides what they use as config, it's also what object they invoke `reload_metadata` on - `self.sp` vs. `self.idp`.

So the proposed change would not work for SP metadata updates - it would be calling `self.idp.reload_metadata`, and `self.idp` would not exist for an SP.

We could also go one step further and add an additional `target` parameter to the generic form (being `self.idp` or `self.sp`), or stick with the original form ...

I initially decided to do two separate methods, after considering what shape a generic one would have to take, but happy to implement that if you'd think that would work out better.

Cheers,
Vlad

c00kiemon5ter:
rebased and merged by b78ab3ee828c4774b83a5bf1a521208c4c2af084 and 206d55d9c951e0c3dc55f4517ae1320059fadead

PR #373, opened 2021-07-07
better endpoint config example for saml2 backend

mheder:
This modification compensates a problematic line here (and the other line for the disco endpoint): 
https://github.com/IdentityPython/SATOSA/blob/84cf60fd38de06a85d6f9d39ef364a3d12571120/src/satosa/backends/saml2.py#L455

If your base_url is like https://www.example.com then this information will be merely thrown away and satosa works.

If your base_url is like https://www.example.com/satosa then you will end up with an endpoint incorrectly registered as satosa/Saml2/acs/post 
Then, the deployment will yield this error:

'Saml2/acs/post' not bound to any function

Which is per se correct as it was registered under another name.
So since the baseurl is thrown away anyway it is better not to have it in the first place.

### All Submissions:

### DIFF
--- a/example/plugins/backends/saml2_backend.yaml.example
+++ b/example/plugins/backends/saml2_backend.yaml.example
@@ -51,9 +51,9 @@ config:
         allow_unsolicited: true
         endpoints:
           assertion_consumer_service:
-          - [<base_url>/<name>/acs/post, 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST']
+          - [/<name>/acs/post, 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST']
           discovery_response:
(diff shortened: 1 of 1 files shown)


mheder:
Actually, what I propose here in turn will result in the wrong binding endpoint locations when you generate the metadata. Therefore there is just no way of running on a baseurl other than a hostname (without path). If you have a <base_url>/<name> in the plugin configuration and the baseurl is not only a hostname, then the plugin will register the wrong endpoint for the ACS, but the metadata will be correct. If you have only /<name>/ then the endpoint gets registered correctly but it will be wrong in the metadata. 

The workaround if you have a base_url with a path element is to have /<name>/ in the plugin config, generate the metadata and fix it by hand later on.

PR #376, opened 2021-08-10
Patch 1

kuba00739:
### All Submissions:

### DIFF
--- a/src/satosa/plugin_loader.py
+++ b/src/satosa/plugin_loader.py
@@ -46,6 +46,14 @@ def load_backends(config, callback, internal_attributes):
         config["BACKEND_MODULES"],
         backend_filter, config["BASE"],
         internal_attributes, callback)
+  
+    for backend in backend_modules:
+        if backend.name in names:
+            raise SATOSAConfigurationError("The name " + backend.name + " is taken!")
+
+        else:
+            names.append(backend.name)
+
     logger.info("Setup backends: {}".format([backend.name for backend in backend_modules]))
     return backend_modules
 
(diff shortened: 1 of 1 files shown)


src/satosa/plugin_loader.py
```diff
@@ -46,6 +46,14 @@ def load_backends(config, callback, internal_attributes):
         config["BACKEND_MODULES"],
         backend_filter, config["BASE"],
         internal_attributes, callback)
+  
+    for backend in backend_modules:
+        if backend.name in names:
+            raise SATOSAConfigurationError("The name " + backend.name + " is taken!")
```
c00kiemon5ter, line 52:
```suggestion
            raise SATOSAConfigurationError("The backend name " + backend.name + " is already used.")
```

src/satosa/plugin_loader.py
```diff
@@ -67,6 +75,13 @@ def load_frontends(config, callback, internal_attributes):
     """
     frontend_modules = _load_plugins(config.get("CUSTOM_PLUGIN_MODULE_PATHS"), config["FRONTEND_MODULES"],
                                      frontend_filter, config["BASE"], internal_attributes, callback)
+    
+    for frontend in frontend_modules:
+        if frontend.name in names:
+            raise SATOSAConfigurationError("The name " + frontend.name + " is taken!")
```
c00kiemon5ter, line 81:
```suggestion
            raise SATOSAConfigurationError("The frontend name " + frontend.name + " is already used.")
```

c00kiemon5ter:
I think it's good to have this.

However, this is doing the right thing but in the wrong way. We should introduce a configuration validation step where this kind of checks should be taking place. Ideally, we can have a configuration schema and check against that directly.

I don't mind merging this to help when these issues come up, but we should re-iterate to properly fix this and other configuration checks as a step in the loading process.

PR #378, opened 2021-08-23
feat: idpy oidcop frontend

peppelinux:
# [oidcop](https://github.com/IdentityPython/oidc-op) Satosa Frontend

- Tested with [JWTConnect-Python-OidcRP](https://github.com/IdentityPython/JWTConnect-Python-OidcRP).
- Mongodb storage for client and session, additional storage engines may be implemented in the future.
- [Demo project](https://github.com/italia/Satosa-Saml2Spid/tree/oidcop)

satosa_oidcop

## endpoints
* [x] provider discovery
* [x] jwks uri
* [x] authorization
* [x] token
* [x] userinfo
* [x] registration
* [x] registration_read endpoint
* [x] introspection endpoint (https://github.com/IdentityPython/SATOSA/pull/378/commits/473310fb5968561f962bf6bcc6b6eacbf78f0b3e)
* [ ] token exchange

## todo
* [x] unit tests
* [x] pytest mongo mock
* [x] test response_type = "code id_token token" (https://github.com/IdentityPython/SATOSA/pull/378/commits/a61dc99503bcb9d4982b77a6ddcf0c41b6732915)
* [x] auto prune expired sessions with mongodb index (https://github.com/IdentityPython/SATOSA/pull/378/commits/137993f77bfb05b44f25ba6df3784e8fb86a31ce, [mongo index](https://github.com/italia/Satosa-Saml2Spid/tree/oidcop/mongo#create-expired-session-deletion))
* [x] token refresh (https://github.com/IdentityPython/SATOSA/pull/378/commits/59c0a53fa73e70551d76c5355c051a7389ab99fd)
* [ ] ~rfc7523 - private_key_jwt test~ > a RP cannot reach the token endpoint if a user have not passed by authz endpoint before. private_key_jwt is a kind of authentication where the user interaction is not needed.
* [ ] DPoP support 
* [ ] general frontend configuration yml file review and refactor (https://github.com/IdentityPython/SATOSA/pull/378/commits/b7668f741b667e0ee8242a234c0380191f50a7df)
* [ ] satosa best practice compliances
* [ ] Refactor scopes/claims configuration following [this](https://github.com/IdentityPython/oidc-op/pull/130/files)

### DIFF
--- a/example/plugins/frontends/oidc_op_frontend.yaml.example
+++ b/example/plugins/frontends/oidc_op_frontend.yaml.example
@@ -4,7 +4,7 @@ name: OIDC
 config:
     domain: &domain localhost
     server_name: *domain
-    base_url: &base_url https://localhost:10000
+    base_url: &base_url <base_url>
 
     storage:
         class: "satosa.frontends.oidcop.storage.mongo.Mongodb"
--- a/example/plugins/frontends/oidc_op_frontend.yaml.example
+++ b/example/plugins/frontends/oidc_op_frontend.yaml.example
@@ -1,0 +1,257 @@
+module: satosa.frontends.idpy_oidcop.OidcOpFrontend
+name: OIDC
+
+config:
+    domain: &domain localhost
+    server_name: *domain
+    base_url: &base_url https://localhost:10000
+
+    storage:
+        class: "satosa.frontends.oidcop.storage.mongo.Mongodb"
+        kwargs:
+            url: mongodb://172.21.0.3:27017/oidcop
+            connection_params:
+                username: satosa
+                password: thatpassword
+                connectTimeoutMS: 5000
+                socketTimeoutMS: 5000
+                serverSelectionTimeoutMS: 5000
+        db_name: oidcop
+        collections:
+            # type: collection name
+            session: session
+            client: client
+
+
+    default_target_backend: spidSaml2
+    salt_size: 8
+
+    op:
+      server_info:
+        add_on:
+          claims:
+            function: oidcop.oidc.add_on.custom_scopes.add_custom_scopes
+            kwargs:
+              research_and_scholarship:
+              - name
+              - given_name
+              - family_name
+              - email
+              - email_verified
+              - sub
+              - iss
+              - eduperson_scoped_affiliation
+          pkce:
+            function: oidcop.oidc.add_on.pkce.add_pkce_support
+            kwargs:
+              code_challenge_method: S256 S384 S512
+              essential: false
+        authentication:
+          user:
+            acr: urn:oasis:names:tc:SAML:2.0:ac:classes:InternetProtocolPassword
+            class: satosa.frontends.oidcop.user_authn.SatosaAuthnMethod
+        authz:
+          class: oidcop.authz.AuthzHandling
+          kwargs:
+            grant_config:
+              expires_in: 43200
+              usage_rules:
+                access_token: {}
+                authorization_code:
+                  max_usage: 1
+                  supports_minting:
+                  - access_token
+                  - refresh_token
+                  - id_token
+                refresh_token:
+                  supports_minting:
+                  - access_token
+                  - refresh_token
+        capabilities:
+          grant_types_supported:
+          - authorization_code
+          - urn:ietf:params:oauth:grant-type:jwt-bearer
+          - refresh_token
+          subject_types_supported:
+          - public
+          - pairwise
+        endpoint:
+          provider_info:
+            class: oidcop.oidc.provider_config.ProviderConfiguration
+            kwargs:
+              client_authn_method: null
+            path: .well-known/openid-configuration
+          authorization:
+            class: oidcop.oidc.authorization.Authorization
+            kwargs:
+              claims_parameter_supported: true
+              client_authn_method: null
+              request_object_encryption_alg_values_supported: &id001
+              - RSA-OAEP
+              - RSA-OAEP-256
+              - A192KW
+              - A256KW
+              - ECDH-ES
+              - ECDH-ES+A128KW
+              - ECDH-ES+A192KW
+              - ECDH-ES+A256KW
+              request_parameter_supported: true
+              request_uri_parameter_supported: true
+              response_modes_supported:
+              - query
+              - fragment
+              - form_post
+              response_types_supported:
+              - code
+            path: OIDC/authorization
+          token:
+            class: oidcop.oidc.token.Token
+            kwargs:
+              client_authn_method:
+              - client_secret_post
+              - client_secret_basic
+              - client_secret_jwt
+              - private_key_jwt
+            path: OIDC/token
+          userinfo:
+            class: oidcop.oidc.userinfo.UserInfo
+            kwargs:
+              claim_types_supported:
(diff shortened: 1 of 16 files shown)


example/plugins/frontends/oidc_op_frontend.yaml.example
```diff
@@ -0,0 +1,257 @@
+module: satosa.frontends.idpy_oidcop.OidcOpFrontend
+name: OIDC
+
+config:
+    domain: &domain localhost
+    server_name: *domain
+    base_url: &base_url https://localhost:10000
```
peppelinux, line 7: this must be taken from proxy general config if I'm not wrong

example/plugins/frontends/oidc_op_frontend.yaml.example
```diff
@@ -0,0 +1,257 @@
...
+            kwargs:
+              client_authn_method:
+              - client_secret_post
+              - client_secret_basic
+              - client_secret_jwt
+              - private_key_jwt
+              release:
+              - username
+            path: OIDC/introspection
+          #registration:
```
peppelinux, line 143: dynamic client registration, registration api and endsession endpoints (logout) was not implemented at this time.

src/satosa/frontends/oidcop/claims.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def combine_join_by_space(values):
+    return " ".join(values)
+
+
+combine_values_by_claim = defaultdict(
+    lambda: combine_return_input,
+    {
+        "sub": combine_select_first_value,
```
peppelinux, line 19: better have this mapping in the frontend configuration, any other additional claim won't take any additional commit to source tree

src/satosa/frontends/oidcop/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class InconsinstentSessionDump(Exception):
```
peppelinux, line 1: to be removed

src/satosa/base.py
```diff
@@ -137,12 +136,18 @@ def _auth_resp_callback_func(self, context, internal_response):
...
-            ]
-            internal_response.subject_id = "".join(subject_id)
+        try:
+            if "user_id_from_attrs" in self.config["INTERNAL_ATTRIBUTES"]:
+                subject_id = [
+                    "".join(internal_response.attributes[attr]) for attr in
+                    self.config["INTERNAL_ATTRIBUTES"]["user_id_from_attrs"]
+                ]
+                internal_response.subject_id = "".join(subject_id)
+        except KeyError as err:
```
peppelinux, line 146: otherwise if a misconfiguration would happen SATOSA returns a KeyError exception from which it is difficult to know the cause from the logs.

This littl change makes satosa just more eloquent

example/plugins/frontends/oidc_op_frontend.yaml.example
peppelinux: https://github.com/IdentityPython/SATOSA/pull/378#

src/satosa/frontends/oidcop/exceptions.py
peppelinux: https://github.com/IdentityPython/SATOSA/pull/378/commits/7faf83572452328df58a503eb61b6b85f89a0106

peppelinux:
Here a brief description about the main characteristics of this implementation.

#### Design principle 1
Anyone can migrate its oidcop configuration, from flask_op or django-oidc-op or whatever, in SATOSA and without any particular efforts. Looking at the [example configuration](https://github.com/peppelinux/SATOSA/blob/oidcop_front/example/plugins/frontends/oidc_op_frontend.yaml.example) we see that `config.op.server_info` have a standard SATOSA configuration with the only addition of the following customizations, needed in SATOSA for interoperational needs. These are:

- autentication
````
        authentication:
          user:
            acr: urn:oasis:names:tc:SAML:2.0:ac:classes:InternetProtocolPassword
            class: satosa.frontends.oidcop.user_authn.SatosaAuthnMethod
````

 - userinfo
 ````
        userinfo:
          class: satosa.frontends.oidcop.user_info.SatosaOidcUserInfo
````

**authentication** inherits `oidcop.user_authn.user.UserAuthnMethod` and overloads two methods involved in user authentication and verification. These task are handled by SATOSA in their authentication backends.

**userinfo** inherits `oidcop.user_info.UserInfo` and proposes a way to store the claims of the users when they comes from the backend. The claims are stored in the session database (actually mongodb) and then they will be fetched during userinfo endpoint (and also token endpoint, for having  them optionally in id_token claims).

#### Weakness of SatosaOidcUserInfo

1. It stores the claims of the user in the db
2. It loads some temporary data and flush them before returning the userinfo response, [here](https://github.com/peppelinux/SATOSA/blob/664bb14ed6879a0f0e59f530241b4d83e112ecd8/src/satosa/frontends/idpy_oidcop.py#L540)

#### unhandled features

- oidcop SSO and cookies were not have been implemented yet because SATOSA doesn't support logout
- dynamic client registration and registration_api are not needed in my deployments but we can have them in a future roadmap
- DPoP2, introspection and exchage SHOULD be implemented

#### Client and Session Storage

MongoDB is the storage, [here](https://github.com/italia/Satosa-Saml2Spid/tree/oidcop/mongo) some brief descriptions for a demo setup. The interface to SATOSA oidcop storage is `satosa.frontends.oidcop.storage.base.SatosaOidcStorage` and it have three methods:

- **get_client_by_id**(self, client_id:str, expired:bool = True)
- **store_session_to_db**(self, session_manager, **kwargs)
- **load_session_from_db**(self, req_args, http_headers, session_manager, **kwargs)

`satosa.frontends.oidcop.storage.mongo.Mongodb` overloads them to have I/O operations on mongodb.

peppelinux:
#### Storage design principles
At this time the storage logic is based on the session_manager load/dump/flush methods.
Each time a request is handled by an endpoint the oidcop session manager flushes its memory and loads the definition from the storage, only which one are strictly related to the request received (client_id, acces_token, 'auth code' ...).

Due to this fact, this [implementation MUST be improved](https://github.com/peppelinux/SATOSA/blob/664bb14ed6879a0f0e59f530241b4d83e112ecd8/src/satosa/frontends/oidcop/storage/mongo.py#L123) and extended to handle the introspection/exchange/registration endpoints.

#### Weakness of the storage model 
it's transactional because it loads data from oidcop's in_memory engine before handling a request and dumps them before doing a response, flushing the inmem storage. Each workers do its own.

each session is in its own, I don't let them grow in concurrency between many workers and requests.
I didn't faced any issues at this time with django-oidc-op and I believe that it would be the same in satosa.

This approach CAN'T be deployd in a multithreaded/asyncio-based approach, but can't see any weakness with standalone workers and indipendent sessions.

example/plugins/frontends/oidc_op_frontend.yaml.example
peppelinux: Implemented!

c00kiemon5ter:
Added to the docs under external contributions: https://github.com/IdentityPython/SATOSA/blob/master/doc/README.md#external-contributions

PR #383, opened 2021-08-30
Make "kid" configurable for the openid_connect frontend

smalihaider-daasi:
SATOSA `OpenIDConnectFrontend` `.well-known/openid-configuration` endpoint has the `jwks_uri` which shows the jwks json, containing a signing key that may be used by the OIDC clients to verify the signature of the id_token. But this jwks, does not currently contain the `kid` element of the jwk.

This pull request adds the `kid` in the jwks if it is configured via the new "signing_key_id" configuration parameter in the openid_connect_frontend.yaml. When initializing `jwkest.jwk.RSAKey` in `OpenIDConnectFrontend`, `kid` needs to be specified, and everything else will automatically be handled by the jwkest python module.

### All Submissions:


c00kiemon5ter:
Thanks @smalihaider-daasi

PR #392, opened 2021-10-26
Fix open redirect in OpenID connect frontend

janpieterk:
## Open redirect in OpenID connect frontend

When running Satosa from the satosa/satosa docker image, the URL:

`http://<your hostname>/<path to your OpenID authorization_endpoint>?response_type=code&client_id=something&scope=openid&redirect_uri=<evil url>`

Redirects to **evil url**.

This seems unwanted behaviour to me, because evil_person can use this open redirect to create a link which points  to a trusworthy domain (**your hostname**) but redirects to a host (**evil url**) under the control of evil_person. This could for example be used in phishing attacks.

I tested this with docker images satosa/satosa:v6.1.0, satosa/satosa:v7.0.3, and satosa/satosa:latest (which at the moment of writing also uses Satosa version 7.0.3).

It seems to me that version 8.0.0 is also vulnerable, but since I did not have an 8.0.0. docker image to test this with, I created a simple fix for the last stable version in use with the satosa/satosa docker image, that is 7.0.3.

This may be too quick and dirty to be merged as-is, but I had to fix our Satosa installations on short notice and this is the best I could come up with. Authorization failures now no longer redirect to external URLs, but show the error on the host where Satosa is running (`BadRequest` instead of `SeeOther`).

Hopefully this will be helpful to you anyway.

### DIFF
--- a/src/satosa/frontends/openid_connect.py
+++ b/src/satosa/frontends/openid_connect.py
@@ -297,11 +297,17 @@ class OpenIDConnectFrontend(FrontendModule):
             logline = lu.LOG_FMT.format(id=lu.get_session_id(context.state), message=msg)
             logger.error(logline)
             error_url = e.to_error_url()
+            if "redirect_uri" in context.request:
+                redirect_uri = context.request["redirect_uri"]
+            else:
+                redirect_uri = False
 
             if error_url:
-                return SeeOther(error_url)
+                return BadRequest("Something went wrong: {}. Not redirecting to external URL {}.".format(str(e), error_url))
+            elif redirect_uri:
+                return BadRequest("Something went wrong: {}. Not redirecting to external URL {}.".format(str(e), redirect_uri))
             else:
-                return BadRequest("Something went wrong: {}".format(str(e)))
+                return BadRequest("Something went wrong: {}.".format(str(e)))
 
         client_id = authn_req["client_id"]
         context.state[self.name] = {"oidc_request": request}


c00kiemon5ter:
hello and thanks for the report! You are right that this should not happen and needs to be fixed. The specification defines that the redirect-uri should be checked against the known redirect-uris of the registered client and if they do not match the resource owner should be informed.

To fix this I've changed the behaviour of the underlying lib to differentiate the error when the redirect-uri is invalid and set `.to_error_url()` to return `None` always, so that nobody can use it. [See, the related diff/commit][pyop_diff]

The effect for the oidc-frontend will be that we will end up on the `BadRequest` case which should have been the initial intention. Users of that frontend should upgrade to [the latest pyop version `v3.3.1`][pyop_v331] (or backport the patch). On the satosa side I've [bumped the minimum pyop version][satosa_pyop_version] required.

Let me know if you have any more issue on this.


  [pyop_diff]: https://github.com/IdentityPython/pyop/compare/687b258febd4c110da2f1c46a34f642ed03d0d07..13ead4a7f3abd3c5b235cb10256131972aa3f277
  [pyop_v331]: https://github.com/IdentityPython/pyop/releases/tag/v3.3.1
  [satosa_pyop_version]: https://github.com/IdentityPython/SATOSA/commit/db4c551c6376a64a9d8b6c0ad6032726e1040163

PR #405, opened 2022-07-20
ModuleRouter: support paths in BASE

### LINKED ISSUE #404: Can't use paths in BASE
I can't make service routing work when base_url contains a path, ie `BASE=https://example.com/satosa`. (I'm trying with an OIDC frontend and SAML2 backend.)

The problem is that both backends register their endpoints with the base path (ie. `satosa/samlbackend/acs/post`), but routing.py expects the backend in the first element of `context.path.split("/")` (routing.py:160). In the example above, the selected backend will be "satosa", which is an invalid backend name.

It doesn't help that `ModuleRouter` doesn't know the base path, so it can not trim the base path from `context.path`.

bajnokk:
If Satosa is installed under a path which is not the root of the
webserver (ie. "https://example.com/satosa"), then endpoint routing must
take the base path into consideration.

Some modules registered some of their endpoints with the base path
included, but other times the base path was omitted, thus it made the
routing fail. Now all endpoint registrations include the base path in
their endpoint map.

Additionally, DEBUG logging was configured for the tests so that the
debug logs are accessible during testing.

Fixes #404 

### All Submissions:

### DIFF
--- a/src/satosa/util.py
+++ b/src/satosa/util.py
@@ -108,7 +108,7 @@ def join_paths(*paths, sep: typing.Optional[str] = None) -> str:
 
     try:
         return leading + sep.join(
-            [path.strip(sep) for path in filter(lambda p: p and p.strip(sep), paths)]
+            path.strip(sep) for path in filter(lambda p: p and p.strip(sep), paths)
         )
     except (AttributeError, TypeError) as err:
         raise TypeError("Arguments must be strings") from err


src/satosa/frontends/openid_connect.py
```diff
@@ -173,6 +173,18 @@ def register_endpoints(self, backend_names):
         :rtype: list[(str, ((satosa.context.Context, Any) -> satosa.response.Response, Any))]
         :raise ValueError: if more than one backend is configured
         """
+        # See https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig
+        # Unfortunately since the issuer is always `base_url` for all OIDC frontend instances,
+        # the discovery endpoint will be the same for every instance.
+        # This means that only one frontend will be usable for autodiscovery.
```
c00kiemon5ter, line 179: > OpenID Providers supporting Discovery MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-configuration` to the Issuer.

The issuer is discovered through a WebFinger request for resources of `http://openid.net/specs/connect/1.0/issuer` relation. The response contains one or more `href` properties with the Issuer URL, which is allowed to contain a path.

What we define is that the frontend contains the frontend name as a path component and under that you can query the well-known documents.

With that in mind we can have multiple frontends each with its own discovery.

The problem is that atm, the `base_url` is used instead of `endpoint_baseurl`.

We can introduce a configuration option to select between the two behaviours, or (even better) introduce a configuration to set the discovery URL for a frontend.

---

At some point I would like to invert this logic; instead of a component defining paths of URLs internally that mapped to functionality (which the routing module has to match to based on some rules), there should be URLs as entrypoints mapped to functionality directly (as it happens within most web frameworks - flask, django, fastapi, etc).

src/satosa/frontends/base.py
```diff
@@ -14,17 +17,22 @@ def __init__(self, auth_req_callback_func, internal_attributes, base_url, name):
...
+        :param base_url: base url of the proxy
         :param name: name of the plugin
         """
         self.auth_req_callback_func = auth_req_callback_func
         self.internal_attributes = internal_attributes
         self.converter = AttributeMapper(internal_attributes)
-        self.base_url = base_url
+        self.base_url = base_url or ""
         self.name = name
+        self.endpoint_baseurl = os.path.join(self.base_url, self.name)
```
c00kiemon5ter, line 34: `os.path.join` will not work on certain platforms (see Windows..).

To join URLs with paths using a function, use `urllib.parse.urljoin`; but it also has caveats (paths that begin with `/` will be considered the root, and bases that do not end with a `/` will be considered a file and will be truncated).

The simplest approach is to just concatenate with a `/` (ie, `f"{self.base_url}/{self.name}"`).

tests/conftest.py
```diff
@@ -38,7 +38,7 @@ def cert_and_key(tmpdir):
 
 @pytest.fixture
 def sp_conf(cert_and_key):
-    sp_base = "http://example.com"
+    sp_base = "http://example.com/test"
```
c00kiemon5ter, line 41: we should parametrize the possible values for the sp and idp and have pytest run the combos, so that we test all cases.

tests/satosa/frontends/test_openid_connect.py
```diff
@@ -369,19 +369,45 @@ def test_jwks(self, context, frontend):
         jwks = json.loads(http_response.message)
         assert jwks == {"keys": [frontend.signing_key.serialize()]}
 
-    def test_register_endpoints_token_and_userinfo_endpoint_is_published_if_necessary(self, frontend):
+    @pytest.mark.parametrize("issuer", [
+        "https://example.com",
+        "https://example.com/some/op",
+        "https://example.com/"
+    ])
+    def test_register_endpoints_handles_path_in_issuer(self, frontend_config, issuer):
```
c00kiemon5ter, line 377: ok, we parametrize here, which I think makes sense ;)

src/satosa/routing.py
```diff
@@ -155,13 +172,12 @@ def endpoint_routing(self, context):
         msg = "Routing path: {path}".format(path=context.path)
         logline = lu.LOG_FMT.format(id=lu.get_session_id(context.state), message=msg)
         logger.debug(logline)
-        path_split = context.path.split("/")
-        backend = path_split[0]
 
-        if backend in self.backends:
+        backend = self._find_backend(context.path)
```
c00kiemon5ter, line 176: there was an idea here to reuse the method that is invoked to find a frontend;

see https://github.com/IdentityPython/SATOSA/pull/279

src/satosa/frontends/openid_connect.py
bajnokk: > The problem is that atm, the base_url is used instead of endpoint_baseurl.
> We can introduce a configuration option to select between the two behaviours, or (even better) introduce a configuration to set the discovery URL for a frontend.

Would you agree to add a `use_module_name_in_issuer` option (default False for backward compatibility, but the examples changed to True)?

A more subtle change but also harder to document alternative would be to make the assignment in [https://github.com/IdentityPython/SATOSA/blob/fc9374c9959ddd57d69ee69f9f683f96ddede964/src/satosa/frontends/openid_connect.py#L66](__init__.py:66)  optional, so that one could set the issuer manually under the `provider` array.

Is any of the two OK with you, or am I misunderstanding the problem?

src/satosa/frontends/base.py
c00kiemon5ter: I would probably name the function `join_url_paths` (verb first).

Note that the latest pysaml2 already requires Python 3.9 and SATOSA will be updated to require it too. IdentityPython projects try to be compatible with the python that ships on the latest Debian stable release (which is now Python 3.9).
So, requiring Python 3.9 is fine; but no newer atm.

src/satosa/frontends/openid_connect.py
c00kiemon5ter: Yes, both are ok with me. As long as we provide a way to configure things to work as before, it is fine to introduce such changes.

src/satosa/routing.py
bajnokk: I must say that I can't fully understand all the `_find_*` methods. My motives for yet another implementation is that it is looking at the very specific part of the request path and tries to find the backend **name**. I'm probably mistaken but I think that the reused version would try to match the backend **endpoint**, which is not present in a frontend request.

src/satosa/frontends/openid_connect.py
bajnokk: The updated patchset contains a commit, which does not perform the above assignment if the provider config has the issuer set. I did some research in git log, but I think it was just in this way forever.
I added a brief explanation to the example configuration, too.
Note that this could be a breaking change for those, who had a lurking "issuer" in their provider config, but since it's never been supported, I'm inclined to go this way rather than adding a new "fix-something-but-dont-break-old-config" type of configuration option.

src/satosa/frontends/base.py
bajnokk: The updated patch contains a `join_paths` implementation, which tries to handle separators a little bit more intelligently than a simple concatenation. I've replaced all erroneous `os.path.join` calls to `join_paths`, but didn't replace `{}/{}.format(foo, bar)` all over the code, since this appears way too many times.

c00kiemon5ter (changes requested):

src/satosa/frontends/openid_connect.py
```diff
@@ -173,6 +173,19 @@ def register_endpoints(self, backend_names):
...
         """
+        # See https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig
+        #
+        # We skip the scheme + host + port of the issuer URL, because we can only map the
+        # path for the provider config endpoint. We are safe to use urlparse().path here,
+        # because for issuer OIDC allows only https URLs without query and fragment parts.
+        issuer = self.provider.configuration_information["issuer"]
+        autoconf_path = ".well-known/openid-configuration"
+        provider_config = (
+            "^{}$".format(join_paths(urlparse(issuer).path.lstrip("/"), autoconf_path)),
```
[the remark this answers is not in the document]
bajnokk: I vaguely remember that I wanted to differentiate between the configuration URL (which must start with the Issuer) and other endpoints (which in theory can be on different URLs), so that BASE=https://url1/path1 and configuration_information["issuer"]=https://url2/path2 would work.
But now I think this is just causing confusion, and I guess a weird setup like this would break anyway and should not be encouraged.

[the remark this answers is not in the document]
bajnokk: Well, giving it some more thinking, I seem to remember why I did this originally.

Both SAML2 frontend and backend supports setting the entity_id to be entirely different from BASE. If I understand it right, also the OpenID backend supports this. I think they do it right, because in a basic flow, the entityID and Issuer are just two identifiers. They get an additional semantic meaning when metadata/provider discovery is in use: then they must be URLs, which point to a downloadable web resource, and hence SATOSA must provide the response.

I think, the current change is more consistent with the rest of the code than it would be if we hard coded BASE there.
What we could do is to support the '<base_url>' string replacement, what I don't really like, but... oh wait, isn't this _already_ supported? (See src/satosa/plugin_loader.py:235. It's surprising, to say at least.)

example/plugins/frontends/openid_connect_frontend.yaml.example
```diff
@@ -33,6 +33,11 @@ config:
   sub_hash_salt: randomSALTvalue
 
   provider:
+    # If you do not specify the issuer here, then BASE will be used as Issuer.
+    # Note that even though this setting must be specified as a full URL,
+    # provider discovery will only work, if the request can be routed back to
+    # SATOSA.
+    issuer: https://op.example.com/satosa/OIDC
```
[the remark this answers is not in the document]
bajnokk: See my reasoning at the discussion above. I want to add <base_url> to the example and a test that this is replaced correctly.

src/satosa/frontends/openid_connect.py
```diff
@@ -173,6 +173,19 @@ def register_endpoints(self, backend_names):
...
+            self.provider_config,
+        )
+        jwks_uri = ("^{}/jwks$".format(self.endpoint_basepath), self.jwks)
```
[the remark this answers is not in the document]
bajnokk: These are not equivalents. The equivalent form would be
`jwks_uri = ("^{}$".format(join_paths("/", self.endpoint_basepath, "jwks")), self.jwks)`
Or is it what you are suggesting that the leading '/' should not be present when endpoint_basepath is the empty string?

src/satosa/frontends/saml2.py
```diff
@@ -675,11 +679,18 @@ def _load_idp_dynamic_endpoints(self, context):
         :param context:
         :return: An idp server
         """
-        target_entity_id = context.target_entity_id_from_path()
+        target_entity_id = self._target_entity_id_from_path(context.path)
         idp_conf_file = self._load_endpoints_to_config(context.target_backend, target_entity_id)
         idp_config = IdPConfig().load(idp_conf_file)
         return Server(config=idp_config)
 
+    def _target_entity_id_from_path(self, request_path):
```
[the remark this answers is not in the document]
bajnokk: `_backend_name_from_path` ?

src/satosa/frontends/openid_connect.py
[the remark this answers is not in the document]
c00kiemon5ter: oh, I didn't think that `self.endpoint_basepath` would be empty.
Previously, `self.name` was used and was always filled; in general, we do not match routes that start with `/`.

So, I still think this is fine, but let me know if I'm skipping over anything.

src/satosa/util.py
```diff
@@ -89,3 +89,29 @@ def rndstr(size=16, alphabet=""):
...
+    try:
+        for p in paths:
+            if not path or path.endswith(sep) or p.startswith(sep):
+                path += p
+            else:
+                path += sep + p
+    except (AttributeError, TypeError) as err:
+        raise TypeError("Arguments must be strings") from err
+
+    return path
```
c00kiemon5ter, line 117: this can be simplified to a join operation with some pre-processing

```suggestion
    sep = "/" if sep is None else sep
    path = sep.join(
        path_stripped
        for path in paths
        for path_stripped in [path.strip(sep)]
        if path_stripped
    )
    if paths and paths[0] and paths[0][0] == sep:
        path = f"{sep}{path}"
    return path
```

c00kiemon5ter (changes requested):
Simplify the join_paths function

src/satosa/util.py
bajnokk: A very similar implementation with additional tests was pushed as a fixup to the original commit which brought in `join_paths`. The only difference is it uses filter instead of the double for, and accounts for `AttributeError`.

PR #407, opened 2022-08-26
oidc_frontend: Add option to mirror public subject

bajnokk:
Add `sub_mirror_subject` configuration parameter. If this is set to
true, the subject received from the backend will be mirrored to the
client, if public sub is used. To maintain backwards compatibility, the
default value is false.

MirrorPublicSubjectIdentifierFactory would normally belong to pyop, but
in order to keep the code and the configuration in the same place, this
code overloads pyop's HashBasedSubjectIdentifierFactory.

### All Submissions:

### DIFF
--- a/src/satosa/frontends/openid_connect.py
+++ b/src/satosa/frontends/openid_connect.py
@@ -46,6 +46,11 @@ from satosa.internal import InternalData
 logger = logging.getLogger(__name__)
 
 
+class MirrorPublicSubjectIdentifierFactory(HashBasedSubjectIdentifierFactory):
+    def create_public_identifier(self, user_id):
+        return user_id
+
+
 class OpenIDConnectFrontend(FrontendModule):
     """
     A OpenID Connect frontend module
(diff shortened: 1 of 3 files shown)


src/satosa/frontends/openid_connect.py
```diff
@@ -460,7 +468,7 @@ def _create_provider(
     return provider
 
 
-def _init_authorization_state(provider_config, db_uri, sub_hash_salt):
+def _init_authorization_state(provider_config, db_uri, sub_hash_salt, mirror_public):
```
c00kiemon5ter, line 471: even though this is a "private" function, it is better to make the new param optional as it defaults to `False` anyway

```suggestion
def _init_authorization_state(provider_config, db_uri, sub_hash_salt, mirror_public=False):
```

c00kiemon5ter (changes requested):
Looks good; just one change to make the new param optional. I think this is good to be merged ;)

src/satosa/frontends/openid_connect.py
bajnokk: Thanks, I've updated the commit. Additionally, I've spotted that there was a lingering assert left in the test code (as a result of a copy-paste).

c00kiemon5ter (approved):

PR #408, opened 2022-09-05
feat: Collect user information when using the Apple backend

melanger:
Apple sends the name only via POST in the first authentication (ever), this change captures it.

### All Submissions:

### DIFF
--- a/src/satosa/backends/apple.py
+++ b/src/satosa/backends/apple.py
@@ -202,13 +202,16 @@ class AppleBackend(BackendModule):
         """
         backend_state = context.state[self.name]
 
-        # Apple sends some user information only via POST in the first request
-        # https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple
+        # Apple has no userinfo endpoint
+        # but may send some user information via POST in the first request.
+        #
+        # References:
+        # - https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple
+        # - https://developer.apple.com/documentation/sign_in_with_apple/namei
         try:
             userdata = context.request.get("user", "{}")
             userinfo = json.load(userdata)
         except Exception as e:
-            # Apple has no userinfo endpoint
             userinfo = {}
 
         authn_resp = self.client.parse_response(
--- a/src/satosa/backends/apple.py
+++ b/src/satosa/backends/apple.py
@@ -203,19 +203,11 @@ class AppleBackend(BackendModule):
         backend_state = context.state[self.name]
 
         # Apple sends some user information only via POST in the first request
-        if "user" in context.request:
-            userinfo = json.load(context.request["user"])
-            userinfo["name"] = " ".join(
-                filter(
-                    None,
-                    [
-                        userinfo.get("firstName", ""),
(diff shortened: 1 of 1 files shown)


src/satosa/backends/apple.py
```diff
@@ -192,24 +201,49 @@ def response_endpoint(self, context, *args):
         :return:
         """
         backend_state = context.state[self.name]
-        authn_resp = self.client.parse_response(AuthorizationResponse, info=context.request, sformat="dict")
+
+        # Apple sends some user information only via POST in the first request
```
c00kiemon5ter, line 205: is this documented?

src/satosa/backends/apple.py
```diff
@@ -192,24 +201,49 @@ def response_endpoint(self, context, *args):
...
+        if "user" in context.request:
+            userinfo = json.load(context.request["user"])
+            userinfo["name"] = " ".join(
```
c00kiemon5ter, line 208: I don't think `name` should be constructed here. This is the job of a micro-service. Here we are only collecting the userinfo and claims with their protocol-specific meaning, as sent by the issuer.

c00kiemon5ter (changes requested):

src/satosa/backends/apple.py
melanger: It's mentioned here (emphasis mine):

> Apple doesn’t receive the user’s full name shared with the system UI. **The raw data is passed directly to your app** from the browser and is not included in the user’s identity token. To help prevent cross-site scripting attacks, validate and sanitize the user-submitted first and last name values before storing on your app servers.

https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple

src/satosa/backends/apple.py
melanger: So you would do this instead?
```py
userinfo = json.load(context.request["user"])
for namePart in ["firstName", "middleName", "lastName"]:
  userinfo[namePart] = userinfo.get(namePart, "")
```

src/satosa/backends/apple.py
```diff
@@ -192,24 +201,41 @@ def response_endpoint(self, context, *args):
...
-        authn_resp = self.client.parse_response(AuthorizationResponse, info=context.request, sformat="dict")
+
+        # Apple sends some user information only via POST in the first request
+        # https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple
+        try:
+            userdata = context.request.get("user", "{}")
+            userinfo = json.load(userdata)
+        except Exception as e:
+            # Apple has no userinfo endpoint
+            userinfo = {}
```
c00kiemon5ter, line 212: small restructure on the inline comments:

```suggestion
        # Apple has no userinfo endpoint
        # but may send some user information via POST in the first request.
        #
        # References:
        # - https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple
        # - https://developer.apple.com/documentation/sign_in_with_apple/namei
        try:
            userdata = context.request.get("user", "{}")
            userinfo = json.load(userdata)
        except Exception as e:
            userinfo = {}
```

c00kiemon5ter (changes requested):

PR #409, opened 2022-09-07
saml2 backend: support using multiple ACS URLs

bajnokk:
When Satosa sends out a SAML2 AuthnRequest, it specifies the AssertionConsumerServiceUrl parameter as well, unless the `hide_assertion_consumer_service` configuration parameter is set. However, Satosa might be deployed in an environment where not all interfaces and host names are accessible for all users. After this change, Satosa tries to select the ACS URL based on the current request, and falls back to the first ACS if there is no match.

### All Submissions:

### DIFF
--- a/src/satosa/backends/saml2.py
+++ b/src/satosa/backends/saml2.py
@@ -298,7 +298,7 @@ class SAMLBackend(BackendModule, SAMLBaseModule):
         except Exception as e:
             msg = "Failed to construct the AuthnRequest for state"
             logline = lu.LOG_FMT.format(id=lu.get_session_id(context.state), message=msg)
-            logger.debug(logline, exc_info=True)
+            logger.error(logline, exc_info=True)
             raise SATOSAAuthenticationError(context.state, "Failed to construct the AuthnRequest") from e
 
         if self.sp.config.getattr('allow_unsolicited', 'sp') is False:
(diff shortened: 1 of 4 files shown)


src/satosa/backends/saml2.py
```diff
@@ -286,10 +286,11 @@ def authn_request(self, context, entity_id):
             kwargs["scoping"] = Scoping(requester_id=[RequesterID(text=requester)])
 
         try:
-            acs_endp, response_binding = self.sp.config.getattr("endpoints", "sp")["assertion_consumer_service"][0]
+            acs_endp, response_binding = self._get_acs(context)
```
c00kiemon5ter, line 289: A different way to organize this is, in terms of strategies which can later become configurable.

So, instead of `_get_acs` that internally tries to match the host and then fallbacks to the first ACS-URL, you can have `get_first_acs`, `get_matching_host_acs`, other-strategies-etc.
Then `_get_acs` becomes

```python
def _get_acs(...):
    result = (
        get_matching_host_acs(...)
        or get_first_acs(...)
    )
    return result
```

---

On the configuration level, a new configuration can be added `resolve_acs_url_strategy` that is a python-path that points to a function. That could be a function already implemented, or a custom function that can be imported. (The strategies need to have the same signature.)

```yaml
resolve_acs_url_strategy: satosa.backends.saml2.get_matching_host_acs
```

and your `_get_acs` would now be

```python
def _get_acs(strategy, ...):
    strategy_fn = _load_fn(strategy) if strategy else get_first_acs
    result = strategy(...)
    return result
```

This allows others to be able to specify their own ways to select an ACS-URL

```yaml
resolve_acs_url_strategy: custom_fns.my_acs_resolver
```

```python
# inside the custom_fns package __init__.py
def my_acs_resolver(...):
    from satosa.backends.saml2 import get_matching_host_acs
    from satosa.backends.saml2 import get_first_acs
    return get_matching_host_acs(...) or get_first_acs(...)
```

bajnokk: I have two problems with specifying the strategy as an importable function name:

1.  `CUSTOM_PLUGIN_MODULE_PATHS` is on the proxy config, and propagating this to the backend level doesn't seem clean to me, but ignoring this is also not very logical
2. To specify the 'matching host' ACS resolver, I'd need to specify `satosa.backends.saml2.get_matching_host_acs`. However, any refactoring of SATOSA source that changes the paths would break existing configurations.

Also, supporting class/static methods also doesn't seem to be obvious, but there might be a clever `importlib` trick that could help me out.

Are you sure that this is worth the effort? What did you have in mind with `_load_fn`?

bajnokk: I've committed a version which makes the ACS selection configurable, but the configuration does not expect a loadable function. I believe that this is cleaner.
Also, the updated code does not change the default behaviour.

bajnokk: Apologies for not formatting the changes with black in the first commit. This makes the second commit more verbose.

c00kiemon5ter: I think this is ok for now. We can reiterate later when we decide how a plugin-like system should load external code.

PR #419, opened 2022-11-04
Handle OIDC prompt together with SAML ForceAuthn and IsPassive

melanger:
Allow processing prompt=none/login/select_account/consent, ForceAuthn=true and IsPassive=true, including interoperability (e.g. IsPassive=true from SAML frontend is converted to prompt=none in OIDC backend)

Change is backward compatible considering nobody is using `prompt` and `IsPassive` on frontends or the constant `Context.KEY_FORCE_AUTHN`.

### All Submissions:

  * Changes are covered by modified existing tests
  * yes, but there were many violations already

### DIFF
--- a/src/satosa/backends/saml2.py
+++ b/src/satosa/backends/saml2.py
@@ -23,7 +23,7 @@ import satosa.util as util
 from satosa.base import SAMLBaseModule
 from satosa.base import SAMLEIDASBaseModule
 from satosa.base import STATE_KEY as STATE_KEY_BASE
-from satosa.context import Context
+from satosa.context import Context, prompt_to_saml_param, get_prompt_list
 from satosa.internal import AuthenticationInformation
 from satosa.internal import InternalData
 from satosa.exception import SATOSAAuthenticationError
@@ -36,14 +36,14 @@ from satosa.metadata_creation.description import (
 )
 from satosa.backends.base import BackendModule
 
-
 logger = logging.getLogger(__name__)
 
 
 def get_memorized_idp(context, config, force_authn):
     memorized_idp = (
-        config.get(SAMLBackend.KEY_MEMORIZE_IDP)
-        and context.state.get(Context.KEY_MEMORIZED_IDP)
+            config.get(SAMLBackend.KEY_MEMORIZE_IDP)
+            and "select_account" not in get_prompt_list(context)
+            and context.state.get(Context.KEY_MEMORIZED_IDP)
     )
     use_when_force_authn = config.get(
         SAMLBackend.KEY_USE_MEMORIZED_IDP_WHEN_FORCE_AUTHN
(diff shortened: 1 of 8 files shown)


src/satosa/backends/saml2.py
```diff
@@ -84,12 +80,13 @@ class SAMLBackend(BackendModule, SAMLBaseModule):
     KEY_SP_CONFIG = 'sp_config'
     KEY_SEND_REQUESTER_ID = 'send_requester_id'
     KEY_MIRROR_FORCE_AUTHN = 'mirror_force_authn'
-    KEY_IS_PASSIVE = 'is_passive'
     KEY_MEMORIZE_IDP = 'memorize_idp'
     KEY_USE_MEMORIZED_IDP_WHEN_FORCE_AUTHN = 'use_memorized_idp_when_force_authn'
 
     VALUE_ACR_COMPARISON_DEFAULT = 'exact'
 
+    SAML_PARAMS = [Context.KEY_SAML_FORCE_AUTHN, Context.KEY_SAML_IS_PASSIVE]
```
c00kiemon5ter, line 88: I don't think it is a good idea to mix the two params.

ForceAuthn is handle in one way.

IsPassive is introduced. It may affect parts of the handling of ForceAuthn but that shouldn't mean they should be bound together.

It is easier to review this if the new functionality is clearly separated from the old.

src/satosa/context.py
```diff
@@ -10,16 +10,56 @@ class SATOSABadContextError(SATOSAError):
...
+def get_prompt_list(context):
+    auth_req_params = get_auth_req_params(context)
+    prompt = auth_req_params.get(Context.KEY_PROMPT, "")
+    return prompt if isinstance(prompt, list) else prompt.split(" ")
+
+
+def prompt_to_saml_param(context, saml_param):
+    prompt_list = get_prompt_list(context)
+    if saml_param == Context.KEY_SAML_IS_PASSIVE:
+        return "none" in prompt_list
```
c00kiemon5ter, line 26: While this may seem reasonable, it is an important assumption that we cannot hardcode.

This should be a configurable choice.

src/satosa/backends/saml2.py
```diff
@@ -50,27 +50,23 @@ def get_memorized_idp(context, config, force_authn):
...
     - the cookie, as it has been stored by the proxy on a redirect to the DS
-      note: the frontend should have been set to mirror the force_authn value.
+      note: the frontend should have been set to mirror the value.
 
     The value is either "true" or None
     """
-    mirror = config.get(SAMLBackend.KEY_MIRROR_FORCE_AUTHN)
-    from_state = mirror and context.state.get(Context.KEY_FORCE_AUTHN)
-    from_context = (
-        mirror and context.get_decoration(Context.KEY_FORCE_AUTHN) in ["true", "1"]
```
c00kiemon5ter, line 69: Keep in mind that "1" is a valid value for booleans for XML. And it is used in practice.

src/satosa/backends/saml2.py
melanger: @c00kiemon5ter They are handled similarly, so I did a little refactoring to prevent adding duplicate code. Do you want to rather split the code (which means duplicity)?

src/satosa/context.py
melanger: Definition of `IsPassive`:

```
If "true", the identity provider and the user agent itself MUST NOT visibly take control
of the user interface from the requester and interact with the presenter in a noticeable fashion.
```

Definition of `prompt=none`:

```
The Authorization Server MUST NOT display any authentication or consent user interface pages. An error is returned if an End-User is not already authenticated or the Client does not have pre-configured consent for the requested Claims or does not fulfill other conditions for processing the request.
```

I believe the meaning is the same, but if you want, I can introduce a configuration option.

melanger: configuration introduced

src/satosa/backends/saml2.py
melanger: restricted to `"true"` and `"1"`

PR #420, opened 2022-11-09
fix: OpenIDConnectFrontend: check for empty db_uri

vladimir-mencl-eresearch:
With IdentityPython/pyop#44 merged, OpenIDConnectFrontend init fails when `db_uri` is not set, as `StorageBase.type` now throws a `ValueError` for db_uri values that do not match one of the recognised storage types (including when `db_uri` is `None`).

Fix this by guarding the `StorageBase.type` with a pythonic test whether `db_uri` was provided.

Same test already guards `StorageBase.from_uri`, add it also to the `StorageBase.type` call made to determine `self.stateless`.

PS: This worked before IdentityPython/pyop#44 was merged because `StorageBase.type` used to _return_  the `ValueError` object, which evaluated as not matching string `"stateless"`.  I also thought about changing `StorageBase.type` to return `None` for `db_uri` of value `None`, but I think it's cleaner to handle it this way and keep the `StorageBase.type` contract clear: Only valid values allowed.

### All Submissions:

* [N/A] Have you written new tests for your changes?
* [N/A] Does your submission pass tests?
* [N/A] This project follows PEP8 style guide. Have you run your code against the 'flake8' linter?

### DIFF
--- a/src/satosa/frontends/openid_connect.py
+++ b/src/satosa/frontends/openid_connect.py
@@ -72,7 +72,7 @@ class OpenIDConnectFrontend(FrontendModule):
         )
 
         db_uri = self.config.get("db_uri")
-        self.stateless = StorageBase.type(db_uri) == "stateless"
+        self.stateless = db_uri and StorageBase.type(db_uri) == "stateless"
         self.user_db = (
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
Sounds good - thank you!

PR #427, opened 2022-11-28
fix AppleBackend

melanger:
common parts are not duplicated

### All Submissions:

### DIFF
--- a/src/satosa/backends/apple.py
+++ b/src/satosa/backends/apple.py
@@ -2,134 +2,20 @@
 Apple backend module.
 """
 import logging
-from datetime import datetime
-from urllib.parse import urlparse
-
+from .openid_connect import OpenIDConnectBackend, STATE_KEY
 from oic.oauth2.message import Message
-from oic import oic
-from oic import rndstr
 from oic.oic.message import AuthorizationResponse
-from oic.oic.message import ProviderConfigurationResponse
-from oic.oic.message import RegistrationRequest
-from oic.utils.authn.authn_context import UNSPECIFIED
-from oic.utils.authn.client import CLIENT_AUTHN_METHOD
-
 import satosa.logging_util as lu
-from satosa.internal import AuthenticationInformation
-from satosa.internal import InternalData
-from .base import BackendModule
-from .oauth import get_metadata_desc_for_oauth_backend
-from ..exception import SATOSAAuthenticationError, SATOSAError
-from ..response import Redirect
-
+from ..exception import SATOSAAuthenticationError
 import json
 import requests
 
 
 logger = logging.getLogger(__name__)
 
-NONCE_KEY = "oidc_nonce"
-STATE_KEY = "oidc_state"
-
 # https://developer.okta.com/blog/2019/06/04/what-the-heck-is-sign-in-with-apple
-class AppleBackend(BackendModule):
+class AppleBackend(OpenIDConnectBackend):
     """Sign in with Apple backend"""
-
-    def __init__(self, auth_callback_func, internal_attributes, config, base_url, name):
-        """
-        Sign in with Apple backend module.
-        :param auth_callback_func: Callback should be called by the module after the authorization
-        in the backend is done.
-        :param internal_attributes: Mapping dictionary between SATOSA internal attribute names and
-        the names returned by underlying IdP's/OP's as well as what attributes the calling SP's and
-        RP's expects namevice.
-        :param config: Configuration parameters for the module.
-        :param base_url: base url of the service
-        :param name: name of the plugin
-
-        :type auth_callback_func:
-        (satosa.context.Context, satosa.internal.InternalData) -> satosa.response.Response
-        :type internal_attributes: dict[string, dict[str, str | list[str]]]
-        :type config: dict[str, dict[str, str] | list[str]]
-        :type base_url: str
-        :type name: str
-        """
-        super().__init__(auth_callback_func, internal_attributes, base_url, name)
-        self.auth_callback_func = auth_callback_func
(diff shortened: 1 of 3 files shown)


melanger:
@c00kiemon5ter Hi, I have a problem and I am not sure how to solve it so that it fits SATOSA logic.

You wrote that a backend should not transform attributes. The problem here (apart from many others) is that "Sign in with Apple" is similar to OIDC, but not fully compliant. Instead of a `name` claim in id_token or userinfo they supply a user dictionary like `'{"name":{"firstName":"Pavel","lastName":"Břoušek"},"email":"anonymousmail@apple.com"}'`.

In the current state, this Apple backend puts the `name` sub-dictionary into `data.attributes`, so I have to have this in `internal_attributes.yml` (because this backend uses the `openid` mappings, not its own):
```
# ...
name:
    openid: [name]
givenname:
    openid: [given_name, name.firstName]
surname:
    openid: [family_name, name.lastName]
# ...
```
This works for Apple backend but fails for standard OIDC providers (e.g. Google), there is an uncaught exception in attribute mapping when it is trying to use the standard `name` claim (string) as a dictionary (because of `name.firstName` in config):

```
[2022-11-29 11:11:21] [ERROR]: [urn:uuid:0eb72947-5f36-421b-a45e-bd515e6050e1] Uncaught exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/satosa/base.py", line 240, in run
    resp = self._run_bound_endpoint(context, spec)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/satosa/base.py", line 180, in _run_bound_endpoint
    return spec(context)
           ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/satosa/backends/openid_connect.py", line 221, in response_endpoint
    internal_resp = self._translate_response(all_user_claims, self.client.authorization_endpoint)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/satosa/backends/openid_connect.py", line 240, in _translate_response
    internal_resp.attributes = self.converter.to_internal("openid", response)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/satosa/attribute_mapping.py", line 100, in to_internal
    attribute_values = self._collate_attribute_values_by_priority_order(external_attribute_name,
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/satosa/attribute_mapping.py", line 119, in _collate_attribute_values_by_priority_order
    attr_val = self._get_nested_attribute_value(attr_name, data)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/satosa/attribute_mapping.py", line 160, in _get_nested_attribute_value
    d = d.get(key)
        ^^^^^
AttributeError: 'str' object has no attribute 'get'
```

I can imagine at least two possible solutions (both of them may be implemented):
1. Change Apple backend so that it does not use `openid` attribute mappings but its own `apple` mappings. Most of them will be duplicated from openid, but the name claims will be different and there will be no problem. This just means changing `self.converter.to_internal("openid", response)` in Apple backend to `self.converter.to_internal("apple", response)`.
2. Change attribute mapping so that it does not fail when trying nested attribute mapping on a non-nested attribute (string).

What do you suggest?

melanger:
@c00kiemon5ter any thoughts on this?

melanger:
@c00kiemon5ter Hi, can you please merge this?

melanger:
@c00kiemon5ter Hi, is there anything blocking this PR? I am using it in production, it works.

src/satosa/backends/apple.py
```diff
@@ -242,78 +110,16 @@ def response_endpoint(self, context, *args):
             raise SATOSAAuthenticationError(context.state, "No user info available.")
 
         all_user_claims = dict(list(userinfo.items()) + list(id_token_claims.items()))
+
+        # convert "string or Boolean" claims to actual booleans
+        for bool_claim_name in ["email_verified", "is_private_email"]:
+            if type(userinfo.get(bool_claim_name)) == str:
+                userinfo[bool_claim_name] = userinfo[bool_claim_name] == "true"
```
c00kiemon5ter, line 117: shouldn't `False` values be set too?

```suggestion
            userinfo[bool_claim_name] = (
                True
                if userinfo[bool_claim_name] == "true"
                else False
            )
```

c00kiemon5ter:
re https://github.com/IdentityPython/SATOSA/pull/427#issuecomment-1330474635

How did you end up with `d` (`data`) being a string? It should have been a dict.

c00kiemon5ter:
OK, I have figured it out; thanks for the comment above.

I have rebased on top of current master and updated the code with 2e3a7823, adding new test cases and fixing the traversal of the internal data.

src/satosa/backends/apple.py
melanger: Those are equivalent. `userinfo[bool_claim_name] == "true"` evaluates to `True` or `False`. But your version might be easier to read.

PR #432, opened 2023-03-15
Filter attributes by shibmd_scope

### LINKED ISSUE #297: SAML2 SP backend check scope on attributes and reject by default
The SAML2 SP backend should check the scope on scoped attributes asserted by an authenticating IdP and compare with the scope from the SAML metadata for the authenticating IdP and by default reject attributes that have the incorrect scope.

Since SATOSA is a proxy and there are some esoteric use cases where the SAML2 SP backend should consume scoped attributes from an authenticating IdP with the incorrect scope, this behavior should be configurable (with reject being the default).

## Code Version

4.x

## Expected Behavior

The SAML2 SP backend should check the scope on scoped attributes asserted by an authenticating IdP and compare with the scope from the SAML metadata for the authenticating IdP and by default reject attributes that have the incorrect scope.

## Current Behavior

The SAML2 SP backend consumes all attributes from the authenticating IdP (provided there is an appropriate definition in internal_attributes.yaml) and makes them available to microservices.

vladimir-mencl-eresearch:
### All Submissions:

Fix #297 by extending `satosa.micro_services.attribute_modifications.FilterAttributeValues` to support not only regexp filtering but also add filter types `shibmdscope_match_scope` and `shibmdscope_match_value` - equivalent to [https://shibboleth.atlassian.net/wiki/spaces/IDP4/pages/1272053894/ScopeMatchesShibMDScope](https://shibboleth.atlassian.net/wiki/spaces/IDP4/pages/1272053894/ScopeMatchesShibMDScope) and [https://shibboleth.atlassian.net/wiki/spaces/IDP4/pages/1272053892/ValueMatchesShibMDScope](https://shibboleth.atlassian.net/wiki/spaces/IDP4/pages/1272053892/ValueMatchesShibMDScope) from the Shibboleth project.

@c00kiemon5ter , please let me know whether you are happy with the overall approach - in #297, you indicated this should be done in a microservice - and in the end, it wasn't that hard.

I first thought I'd do this in a new microservice, but I'd just be replicating what `FilterAttributeValues` does - and this additional filtering fits in there conceptually well.

If happy with this approach, I'd add documentation and tests - but please let me know if overall happy with the approach I took.

Cheers,
Vlad

### DIFF
--- a/example/plugins/microservices/filter_attributes.yaml.example
+++ b/example/plugins/microservices/filter_attributes.yaml.example
@@ -17,6 +17,16 @@ config:
                 eduPersonPrincipalName:
                     # enforce correct scope
                     shibmdscope_match_scope:
+                subject-id:
+                    # enforce attribute syntax
+                    regexp: "^[0-9A-Za-z][-=0-9A-Za-z]{0,126}@[0-9A-Za-z][-.0-9A-Za-z]{0,126}\\Z"
+                    # enforce correct scope
+                    shibmdscope_match_scope:
+                pairwise-id:
+                    # enforce attribute syntax
+                    regexp: "^[0-9A-Za-z][-=0-9A-Za-z]{0,126}@[0-9A-Za-z][-.0-9A-Za-z]{0,126}\\Z"
+                    # enforce correct scope
+                    shibmdscope_match_scope:
                 schacHomeOrganization:
                     # enforce scoping rule on attribute value
                     shibmdscope_match_value:
--- a/example/plugins/microservices/filter_attributes.yaml.example
+++ b/example/plugins/microservices/filter_attributes.yaml.example
@@ -2,6 +2,15 @@ module: satosa.micro_services.attribute_modifications.FilterAttributeValues
 name: AttributeFilter
 config:
     attribute_filters:
+        # default rules for any IdentityProvider
+        "":
+            # default rules for any requester
+            "":
+                # enforce controlled vocabulary
+                eduPersonAffiliation: "^(faculty|student|staff|alum|member|affiliate|employee|library-walk-in)$"
+                eduPersonPrimaryAffiliation: "^(faculty|student|staff|alum|member|affiliate|employee|library-walk-in)$"
+                eduPersonScopedAffiliation: "^(faculty|student|staff|alum|member|affiliate|employee|library-walk-in)@"
+
         target_provider1:
             requester1:
                 attr1: "^foo:bar$"
--- a/example/plugins/microservices/filter_attributes.yaml.example
+++ b/example/plugins/microservices/filter_attributes.yaml.example
@@ -6,10 +6,20 @@ config:
         "":
             # default rules for any requester
             "":
-                # enforce controlled vocabulary
+                # enforce controlled vocabulary (via simple notation)
                 eduPersonAffiliation: "^(faculty|student|staff|alum|member|affiliate|employee|library-walk-in)$"
                 eduPersonPrimaryAffiliation: "^(faculty|student|staff|alum|member|affiliate|employee|library-walk-in)$"
-                eduPersonScopedAffiliation: "^(faculty|student|staff|alum|member|affiliate|employee|library-walk-in)@"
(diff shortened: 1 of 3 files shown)


example/plugins/microservices/filter_attributes.yaml.example
```diff
@@ -2,6 +2,25 @@ module: satosa.micro_services.attribute_modifications.FilterAttributeValues
...
+            # default rules for any requester
+            "":
+                # enforce controlled vocabulary (via simple notation)
+                eduPersonAffiliation: "^(faculty|student|staff|alum|member|affiliate|employee|library-walk-in)$"
+                eduPersonPrimaryAffiliation: "^(faculty|student|staff|alum|member|affiliate|employee|library-walk-in)$"
+                eduPersonScopedAffiliation:
+                    # enforce controlled vocabulary (via extended notation)
+                    regexp: "^(faculty|student|staff|alum|member|affiliate|employee|library-walk-in)@"
+                    # enforce correct scope
+                    shibmdscope_match_scope:
```
c00kiemon5ter, line 16: These filters stand here without a key.
Wouldn't it be more natural to turn them into booleans?

The code would do:
```py
                elif filter_type == "shibmdscope_match_scope" and filter_value:
```

I know this is against the implementation logic (a loop over the filters) but I think from the point of view of a user it feels better.

Let me know.

src/satosa/micro_services/attribute_modifications.py
```diff
@@ -29,28 +33,62 @@ def __init__(self, config, *args, **kwargs):
...
+                elif attribute_name in attributes:
+                    attributes[attribute_name] = list(filter(filter_func, attributes[attribute_name]))
+
+    def _shibmdscope_match_value(self, value, md_scopes):
+        for md_scope in md_scopes:
+            if not md_scope['regexp'] and md_scope['text'] == value:
+                return True
+            elif md_scope['regexp'] and re.compile(md_scope['text']).match(value):
+                return True
+        return False
```
c00kiemon5ter, line 86: Compiling the regex will not make a difference here, as the result is not reused;
I'm thinking if it would make sense to use `re.fullmatch` here, but I guess `re.match` is safer.

c00kiemon5ter:
@vladimir-mencl-eresearch I think it's fine! Please, go ahead 👍

example/plugins/microservices/filter_attributes.yaml.example
vladimir-mencl-eresearch: I was considering several options for how the filters would be specified.

I wanted to:
* allow selecting the filter type (going beyond just a regexp)
* allow using multiple filters for a single attribute (eduPersonScopedAffiliation needs both the md_scope filter and a regexp filter for the value).

I considered a list - like:
```
eduPersonScopedAffiliation:
  - type: regexp
    text: staff|student
  - type: shibmdscope_match_scope
```
... but thought that would be too verbose - and the list should not be required.

So went for a dict indexed by the filter type - and meaning of value is specific to the filter type.  And some filter types do not need a value at all.

So in the syntax I picked, it is actually a key with no value (from YAML parser, with value `None`).

I think I got influenced by Ansible when picking that - Ansible uses that a lot.  Ansible tasks get specified by including a dict key matching the name of the task - and if the task needs do parameters, you just have a key with no value - like here.

I could change that - but adding any value for these keys would just make the syntax more verbose, so having just the keys seems the most concise way to me.

Your thoughts?  (Or preferences on syntax?)

vladimir-mencl-eresearch:
Hi @c00kiemon5ter , I've  now added tests - and also made a few tweaks, incl. removing the unnecessary re.compile - and switched to re.fullmatch as discussed.

I think it's now ready for a proper review - please let me know what you think needs to get done to get it merged.

Thanks a lot in advance for getting back to me.

Cheers,
Vlad

example/plugins/microservices/filter_attributes.yaml.example
c00kiemon5ter: OK, let's keep it as it is now.

PR #433, opened 2023-03-21
backends/saml2: gracefully handle mismatching ACS

bajnokk:
This is a new attempt to solve #324 . The error message specifically targets the administrator / logs. Even though the exception message has an assumption (the original relay state might be missing for other, more unlikely reasons), but drawing the admin's attention to the mismatching request address and ACS address is a good idea in this case, IMHO.

Commit message follows:

When the IdP redirects to an ACS which has a different address than the one we used for initiating the request, we are unable to verify the RelayState, since the browser does not send the session cookie. In order to make configuration debugging easier, raise an explanatory SATOSAAuthenticationError instead of a KeyError.

While adding a unit test to check for the proper error reporting, some code duplication was refactored.

### All Submissions:

### DIFF
--- a/src/satosa/backends/saml2.py
+++ b/src/satosa/backends/saml2.py
@@ -404,11 +404,18 @@ class SAMLBackend(BackendModule, SAMLBaseModule):
             del self.outstanding_queries[req_id]
 
         # check if the relay_state matches the cookie state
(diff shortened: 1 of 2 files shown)


c00kiemon5ter:
closed by d9864643a5c606b2e2da14c0f5d6c5271178b5ac and 62f8775421734af08a337be18ff208d00a78bc71

PR #439, opened 2023-06-26
Add new OIDC backend based on idpy-oidc

rohe:
### All Submissions:

This is an OIDC/OAuth2 backend based on the [idpy-oidc](https://github.com/IdentityPython/idpy-oidc) package.
Others will supply the OIDC/OAuth2 frontend.

### DIFF
--- a/src/satosa/backends/idpy_oidc.py
+++ b/src/satosa/backends/idpy_oidc.py
@@ -1,0 +1,124 @@
+"""
+OIDC backend module.
+"""
+import logging
+from datetime import datetime
+
+from idpyoidc.server.user_authn.authn_context import UNSPECIFIED
+
+from satosa.backends.base import BackendModule
+from satosa.internal import AuthenticationInformation
+from satosa.internal import InternalData
+
+logger = logging.getLogger(__name__)
+
+"""
+OIDC/OAuth2 backend module.
+"""
+from idpyoidc.client.oauth2.stand_alone_client import StandAloneClient
+
+
+class IdpyOIDCBackend(BackendModule):
+    """
+    Backend module for OIDC and OAuth 2.0, can be directly used.
+    """
+
+    def __init__(self,
+                 outgoing,
+                 internal_attributes,
+                 config,
+                 base_url,
+                 name,
+                 external_type,
+                 user_id_attr
+                 ):
+        """
+        :param outgoing: Callback should be called by the module after the authorization in the
+        backend is done.
+        :param internal_attributes: Mapping dictionary between SATOSA internal attribute names and
+        the names returned by underlying IdP's/OP's as well as what attributes the calling SP's and
+        RP's expects namevice.
+        :param config: Configuration parameters for the module.
+        :param base_url: base url of the service
+        :param name: name of the plugin
+        :param external_type: The name for this module in the internal attributes.
+
+        :type outgoing:
+        (satosa.context.Context, satosa.internal.InternalData) -> satosa.response.Response
+        :type internal_attributes: dict[string, dict[str, str | list[str]]]
+        :type config: dict[str, dict[str, str] | list[str]]
+        :type base_url: str
+        :type name: str
+        :type external_type: str
+        """
+        super().__init__(outgoing, internal_attributes, base_url, name)
+        self.name = name
+        self.external_type = external_type
+        self.user_id_attr = user_id_attr
+
+        self.client = StandAloneClient(config=config["client_config"],
+                                       client_type=config["client_config"]['client_type'])
+        # Deal with provider discovery and client registration
+        self.client.do_provider_info()
+        self.client.do_client_registration()
+
+    def start_auth(self, context, internal_request):
+        """
+        See super class method satosa.backends.base#start_auth
+
+        :type context: satosa.context.Context
+        :type internal_request: satosa.internal.InternalData
+        :rtype satosa.response.Redirect
+        """
+        return self.client.init_authorization()
+
+    def register_endpoints(self):
+        """
+        Creates a list of all the endpoints this backend module needs to listen to. In this case
+        it's the authentication response from the underlying OP that is redirected from the OP to
+        the proxy.
+        :rtype: Sequence[(str, Callable[[satosa.context.Context], satosa.response.Response]]
+        :return: A list that can be used to map the request to SATOSA to this endpoint.
+        """
+
+        return self.client.context.claims.get_usage('authorization_endpoint')
+
+    def _authn_response(self, context):
+        """
+        Handles the authentication response from the AS.
+
+        :type context: satosa.context.Context
+        :rtype: satosa.response.Response
+        :param context: The context in SATOSA
+        :return: A SATOSA response. This method is only responsible to call the callback function
+        which generates the Response object.
+        """
+
+        _info = self.client.finalize(context.request)
+
+        try:
+            auth_info = self.auth_info(context.request)
+        except NotImplementedError:
+            auth_info = AuthenticationInformation(UNSPECIFIED, str(datetime.now()), _info["issuer"])
+
+        internal_response = InternalData(auth_info=auth_info)
+        internal_response.attributes = self.converter.to_internal(self.external_type,
+                                                                  _info['userinfo'])
+        internal_response.subject_id = _info['userinfo'][self.user_id_attr]
+        del context.state[self.name]
+        # return self.auth_callback_func(context, internal_response)
+        if 'error' in _info:
+            return _info
+        else:
+            return _info['userinfo']
+
+    def auth_info(self, request):
+        """
+        Creates the SATOSA authentication information object.
+        :type request: dict[str, str]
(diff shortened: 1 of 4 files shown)


src/satosa/exception.py
```diff
@@ -68,7 +68,9 @@ def message(self):
         """
         return self._message.format(error_id=self.error_id)
 
-
+class UnknownSystemEntity(SATOSAError):
```
c00kiemon5ter, line 71: I can already tell that is going to be useful for making the saml bits optional, but in this changeset this exception is not used.

src/satosa/backends/idpy_oidc.py
```diff
@@ -0,0 +1,144 @@
...
+        :type issuer: str
+        :type subject_type: str
+        :rtype: InternalData
+
+        :param response: Dictioary with attribute name as key.
+        :param issuer: The oidc op that gave the repsonse.
+        :param subject_type: public or pairwise according to oidc standard.
+        :return: A SATOSA internal response.
+        """
+        auth_info = AuthenticationInformation(UNSPECIFIED, str(datetime.now()), issuer)
```
c00kiemon5ter, line 123: We used to use `datetime.now` in the old module, but I think we should set the the `auth_time` of the id-token. 

Similarly, for the ACCR we should set `acr` or `amr` (not sure which of the two fits better) if present.

```suggestion
        auth_class_ref = response["auth_time"]
        timestamp = response.get("amr", UNSPECIFIED)
        auth_info = AuthenticationInformation(auth_class_ref, timestamp, issuer)
```

src/satosa/backends/idpy_oidc.py
```diff
@@ -0,0 +1,144 @@
...
+        userinfo = _info.get('userinfo')
+        id_token = _info.get('id_token')
+
+        if not id_token and not userinfo:
+            msg = "No id_token or userinfo, nothing to do.."
+            logline = lu.LOG_FMT.format(id=lu.get_session_id(context.state), message=msg)
+            logger.error(logline)
+            raise SATOSAAuthenticationError(context.state, "No user info available.")
+
+        all_user_claims = dict(list(userinfo.items()) + list(id_token.items()))
```
c00kiemon5ter, line 103: Could it happen that the items are conflicting?
eg, both id-token and userinfo returning an `email` claim.
Which one wins when the values differ?

src/satosa/backends/idpy_oidc.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def start_auth(self, context, internal_request):
+        """
+        See super class method satosa.backends.base#start_auth
+
+        :type context: satosa.context.Context
+        :type internal_request: satosa.internal.InternalData
+        :rtype satosa.response.Redirect
+        """
+        login_url = self.client.init_authorization()
```
c00kiemon5ter, line 61: The login_url will be the same always. Then this should be calculated once during initialization (`__init__`).

src/satosa/backends/idpy_oidc.py
```diff
@@ -0,0 +1,144 @@
...
+        Creates a list of all the endpoints this backend module needs to listen to. In this case
+        it's the authentication response from the underlying OP that is redirected from the OP to
+        the proxy.
+        :rtype: Sequence[(str, Callable[[satosa.context.Context], satosa.response.Response]]
+        :return: A list that can be used to map the request to SATOSA to this endpoint.
+        """
+        url_map = []
+        redirect_path = self.client.context.claims.get_usage('redirect_uris')
+        if not redirect_path:
+            raise SATOSAError("Missing path in redirect uri")
```
c00kiemon5ter, line 75: Then this check could move to the initialization (`__init__`).

c00kiemon5ter (changes requested):

src/satosa/backends/idpy_oidc.py
rohe: The attribute names confuses me completely.

`auth_time` in the id-token has nothing to do with `acr` or `amp` so I don't understand this change.

Shouldn't it be something like:

```
timestamp = response["auth_time"]
auth_class_ref = response.get("amr", response.get("acr", UNPECIFIED))
auth_info = AuthenticationInformation(auth_class_ref, timestamp, issuer)
```

src/satosa/exception.py
rohe: OK

src/satosa/backends/idpy_oidc.py
rohe: It could definitely happen. The id_token values would override.

src/satosa/backends/idpy_oidc.py
rohe: As the line looks like right now that is true.
I don't know if context and/or internal request which are both arguments to the call could have any influence on what the authorisation request should look like.

As it's written right now that only things that matter is client configuration and the provider information and if this is how it should be then you're right.

src/satosa/backends/idpy_oidc.py
rohe: Yes

src/satosa/backends/idpy_oidc.py
rohe: Or perhaps 
`auth_class_ref = response.get("acr", response.get("amr", UNPECIFIED))`
is a better order

src/satosa/backends/idpy_oidc.py
rohe: The id_token values should override. Since that information is in a signed JWT. So it's protected.
The userinfo information may be signed and/or encrypted. Though mostly it will be unprotected.

The expression may be rewritten to use `.update()`.
The benefit from using concatenation and the conversion to dictionary is that neither **userinfo** nor **id_token** is modified. Otherwise `.update()` is faster and using less memory.

c00kiemon5ter: OK, so let's keep this as it is for now.

src/satosa/backends/idpy_oidc.py
```diff
@@ -0,0 +1,148 @@
...
+        :type subject_type: str
+        :rtype: InternalData
+
+        :param response: Dictioary with attribute name as key.
+        :param issuer: The oidc op that gave the repsonse.
+        :param subject_type: public or pairwise according to oidc standard.
+        :return: A SATOSA internal response.
+        """
+        timestamp = response["auth_time"]
+        auth_class_ref = response.get("amr", response.get("acr", UNSPECIFIED))
```
c00kiemon5ter, line 125:
```suggestion
        auth_class_ref = response.get("amr") or response.get("acr") or UNSPECIFIED
```

c00kiemon5ter:
This needs the stand-alone client to be merged (https://github.com/IdentityPython/idpy-oidc/pull/70) before this is merged.

src/satosa/backends/idpy_oidc.py
```diff
@@ -0,0 +1,148 @@
...
+        :type issuer: str
+        :type subject_type: str
+        :rtype: InternalData
+
+        :param response: Dictioary with attribute name as key.
+        :param issuer: The oidc op that gave the repsonse.
+        :param subject_type: public or pairwise according to oidc standard.
+        :return: A SATOSA internal response.
+        """
+        timestamp = response["auth_time"] or datetime.utcnow()
```
smalihaider, line 124: response["auth_time"] may lead to KeyError if auth_time is missing from the response.

Moreover, datatime.utcnow() would return datetime object which is not serializable. I think we need to change it to something as follows:

`
timestamp = response.get("auth_time")
 if not timestamp:
     timestamp = datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
`

c00kiemon5ter: `auth_time` is required to be there.

The datetime should be a timestamp
```suggestion
        timestamp = response["auth_time"] or int(datetime.utcnow().timestamp())
```

smalihaider: I tried to authenticate via Microsoft AD and got KeyError at this line and that's why I thought may be response would not always have `auth_time`. Do you think it's an issue with MS AD? Nonetheless, I think your following suggestion:

`timestamp = response["auth_time"] or int(datetime.utcnow().timestamp())
`
should be adopted.

c00kiemon5ter: I am sorry, I went back to the spec to check and indeed `auth_time` is not required. `iat` is required, so that would be a better choice. So probably just have

```suggestion
        timestamp = response["iat"]
```

but I think it is also ok to have

```suggestion
        timestamp = (
            response.get("auth_time")
            or response.get("iat")
            or int(datetime.utcnow().timestamp()
        )
```

smalihaider: Yes, I agree @c00kiemon5ter  - thanks.

rohe: We should put the right value here. I'll look into this.
Probably tomorrow.

PR #440, opened 2023-07-17
feat: process multiple values of the resource query param

kofzera:
Changed the parsing of the query parameters so that in case of more than one occurrence of the same parameter, the value in the dictionary gets converted to a list, and additional values are appended. This allows sending an array value in the queries (useful, for example, with the "resource" parameter in the authorization requests). 


c00kiemon5ter:
This will make things very hard to work with. Wherever a query param is used, we would need to check if it is a list or not..

In general, with both OIDC and SAML we do not expect to get multiple query params. The Resource Indicators specification breaks this rule.

Instead of making this a generic processing step, I think that we should check for the `resource` query param(s) explicitly and set it as a list, always.

PR #446, opened 2023-11-08
fix: prevent endless loop from accessing self

thomaswar:
Fix to prevent endless loop when `self.data` calls `self.__getattr__`

### All Submissions:

### DIFF
--- a/src/satosa/internal.py
+++ b/src/satosa/internal.py
@@ -35,7 +35,7 @@ class _Datafy(UserDict):
 
     def __getattr__(self, key):
         if key == "data":
-            return self.data
+            return super().data
 
         try:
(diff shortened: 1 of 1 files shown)


thomaswar:
Hello @c00kiemon5ter, 

I hit this issue with a custom modification that needs to serialize and unserialize the state to implement some special flows and thus tampers with data.

I think that particular code is just not reached otherwise yet. I am sure accessing self.data this way should have always ended with a maximum recursion error. I am not aware of any history of UserDict that would have diverted this access to super() by itself.

c00kiemon5ter:
I don't seem to be able to trigger this, but I think it's good in any case.

PR #447, opened 2023-11-13
fix: do not access missing props in Apple backend

melanger:
Introduced in #427, sorry @c00kiemon5ter I have missed that you have removed the guarding condition. The properties are not always present and might be already boolean, in which case the previous code converts `False` to `True`.

### All Submissions:

### DIFF
--- a/src/satosa/backends/apple.py
+++ b/src/satosa/backends/apple.py
@@ -113,11 +113,10 @@ class AppleBackend(OpenIDConnectBackend):
 
         # convert "string or Boolean" claims to actual booleans
         for bool_claim_name in ["email_verified", "is_private_email"]:
-            userinfo[bool_claim_name] = (
-                True
-                if userinfo[bool_claim_name] == "true"
-                else False
-            )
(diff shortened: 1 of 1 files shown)


src/satosa/backends/apple.py
```diff
@@ -113,11 +113,10 @@ def response_endpoint(self, context, *args):
 
         # convert "string or Boolean" claims to actual booleans
         for bool_claim_name in ["email_verified", "is_private_email"]:
-            userinfo[bool_claim_name] = (
-                True
-                if userinfo[bool_claim_name] == "true"
-                else False
-            )
+            if type(all_user_claims.get(bool_claim_name)) == str:
```
c00kiemon5ter, line 116: do we care if it is a str,
or whether the claim is there?

```suggestion
            if bool_claim_name in all_user_claims.keys():
```

melanger: Yes, because the claim also might be a boolean, according to Apple's spec, and it there was `True` as boolean, your code woud turn it into `False`, because `False != "true"`

PR #449, opened 2023-12-19
Handle multiple back/front-ends

theseal:
Without this fix only the last back/front-end will be written to file if split is not involved.

Add new method create_entities_descriptor as a counterpart to create_signed_entity_descriptor to also apply `valid` option to EntititesDescriptor but avoiding signing.

### All Submissions:

### DIFF
--- a/src/satosa/metadata_creation/saml_metadata.py
+++ b/src/satosa/metadata_creation/saml_metadata.py
@@ -169,3 +169,19 @@ def create_entity_descriptor_metadata(entity_descriptor, valid_for=None):
         entity_descriptor.valid_until = in_a_while(hours=valid_for)
 
     return str(entity_descriptor)
+
+def create_entities_descriptor(entity_descriptors, valid_for=None):
+    """
+    :param entity_descriptors: the entity descriptors to put in in an EntitiesDescriptor
+    :param valid_for: number of hours the metadata should be valid
+    :return: the EntitiesDescriptor metadata
+
+    :type entity_descriptors: Sequence[saml2.md.EntityDescriptor]]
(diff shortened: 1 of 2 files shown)


vladimir-mencl-eresearch:
Hi, 

Looking at this PR, I understand this would introduce a breaking change: even for deployments with just a entity entity descriptor, the `EntityDescriptor` would now be wrapped in an `EntitiesDescriptor`.  This might take many deployments by surprise - break the unwritten assumption that the file would contain just a single `EntityDescriptor` element.

Though, I'm not sure what the best way forward would be:
* doing this conditional on number of `length(entity_descriptors) > 1` would be more likely to cause issues
* maybe a new option to wrap multiple descriptors into an EntitiesDescriptor?
* or accept this PR and flag it as a breaking change

Note that I'm not a maintainer, so these are just my thoughts, not a maintainer review.

Hope its helpful.

Cheers,
Vlad

theseal:
You are right and thought about it when I started to write the change but apparently forgot about it.

We have to wonder what the intent was when the code was written from the begging I guess. Stacking multiple entityDescriptor's without embracing in an entitiesDescriptor I think is incorrect.

Not sure how common it is to have multiple front/back-ends (we have at-least) so to mitigate the break I could add your suggested check for the amount of entities and only wrap multiple entities in an entitiesDescriptor.

vladimir-mencl-eresearch:
I have a use case where I can have an arbitrary number of SAML backends (each service bridged has a standalone SAML identity), but I use the `--split-backend` to get each `EntityDescriptor` in a separate metadata file.

PR #461, opened 2024-04-08
Add an EntraID OIDC backend based on the included OpenIDConnectBackend

ceko:
# EntraID OIDC Backend

We have a need to use EntraID as an authentication backend, but EntraID does not closely follow the OpenID spec and can't be integrated with the current backends. In particular, the issuer used in discovery does not always match the issuer returned in other parts of the login flow.

I built a backend using the OpenIDConnectBackend as a model and integrated with [the msal library.](https://github.com/AzureAD/microsoft-authentication-library-for-python) I also wrote several tests using the same backend's tests as a model for coverage.

Please let me know if this is something you'd like to include in the base product, if not I'll continue using it as a module in my SATOSA instance.

### All Submissions:

### DIFF
--- a/example/plugins/backends/entraid_oidc_backend.yaml.example
+++ b/example/plugins/backends/entraid_oidc_backend.yaml.example
@@ -1,0 +1,41 @@
+module: satosa.backends.entraid_oidc.EntraIDOIDCBackend
+name: entraid_oidc
+config:
+  # https://learn.microsoft.com/en-us/entra/identity-platform/v2-protocols-oidc#find-your-apps-openid-configuration-document-uri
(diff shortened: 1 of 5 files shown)


c00kiemon5ter:
Thank you for this work @ceko and thank for making the dependencies optional.

I think that we can pull this in. If we need to make changes and to maintain this module, would you want to be contacted to take care of this?

ceko:
I don't _need_ to be contacted, but I wouldn't mind being marked as a maintainer for this backend. Would this change make it into the Docker image as well? We are currently using the image for an upcoming project, I'd like to prepare for having/not having the msal library.

PR #462, opened 2024-04-12
add WSGIPassAuthorization directive to Apache configuration

guillomovitch:
I spend a few hours debugging this token endpoint authentification issue, which is espcially painful as only basic authentifcation is supported.

For the record, here are the following symptoms:
```
[satosa.frontends.openid_connect.token_endpoint] invalid client authentication at token endpoint
...
pyop.exceptions.InvalidClientAuthentication: client_id 'None' unknown
```


guillomovitch:
Hello. May I have any kind of feedback about this PR, wich seems to be relatively trivial ?

PR #464, opened 2024-04-25
BaseProcessor: add missing 'self'

jonathanperret:
The `BaseProcessor.process` method is missing its `self` argument. I stumbled upon this while writing a custom attribute processor because pylint complained that I was overriding a method with a differing number of arguments.


c00kiemon5ter:
thanks for the catch ;)

Issue #179, opened 2018-03-06
BASE must not include a path in mod_wsgi-hosted deployments


xenophonf:
## Code Version
master

## Expected Behavior
I should be able to host SATOSA at any valid URL path, e.g., setting **BASE** to `https://federation.example.com/satosa`.

Likewise, the following values of **BASE** should be treated the same: `https://federation.example.com/` and `https://federation.example.com`.  However, only the latter value of BASE results in a working configuration.

## Current Behavior
If one hosts the SATOSA WSGI application at a path other than /, and if one sets **BASE** in the SATOSA proxy configuration file to something other than `<scheme>://<host>`, then the SAML back end registers invalid endpoint mappings, e.g. with **BASE** set to `https://federation.example.com/satosa`, it sets up these mappings (**url_map**):

```python
[('^satosa/Saml2/acs/post$', functools.partial(<bound method SAMLBackend.authn_response of <satosa.backends.saml2.SAMLBackend object at 0x7fcfc24a61d0>>, binding='urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST')), ('^satosa/Saml2/acs/redirect$', functools.partial(<bound method SAMLBackend.authn_response of <satosa.backends.saml2.SAMLBackend object at 0x7fcfc24a61d0>>, binding='urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect'))]
```

This results in the following error:

```
Unable to determine the entityID's for the IdP or SP
```

Whereas with **BASE** set to `https://federation.example.com`, it sets up these mappings

```python
[('^/Saml2/acs/post$', functools.partial(<bound method SAMLBackend.authn_response of <satosa.backends.saml2.SAMLBackend object at 0x7f89c4aff1d0>>, binding='urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST')), ('^/Saml2/acs/redirect$', functools.partial(<bound method SAMLBackend.authn_response of <satosa.backends.saml2.SAMLBackend object at 0x7f89c4aff1d0>>, binding='urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect'))]
```

Due to differences in how the SAML front end configures endpoint mappings, it works as expected for both values of **BASE**.

## Possible Solution
This might be related to issue #148.

As a workaround for situations where one might wish to co-locate SATOSA and other web applications or content on the same virtual host, use the `WSGIScriptAliasMatch` directive with a negative lookahead:

```
[...]

Issue #148, opened 2017-09-15
Inconsistent use of base_url throughout the code


mrvanes:
I tried to map SATOSA to a URI'd path behind a reverse proxy and didn't succeed. Looking through the code there are many mixes of the use of BASE (self.base_url) that defeat using it properly.

1. metadata (generation) uses it
2. register_endpoints of saml front and backend modules don't use it, but oidc frontend does
3. discovery return_url doesn't use it
4. saml2 backend ACS doesn't

This is where I stopped debugging and fixing the code.

All the above problems can be overcome seperately by wisely choosing a combination of reverse proxy conf (include base path in backend request or not) and SATOSA conf (with or without base path), but never together as as working system. It's a classical waterbed.

This issue is here only to address the problem, because we decided to go vhost based reverse proxy to keep the project going.

To clarify what I mean with base path URI'd SATOSA:
reverse proxy address: https://<host>/satosa/
backend server: http://localhost:8080/ or http://localhost:8080/satosa

PR #483, opened 2025-01-21
Fix SAML2 endpoints registration when base_url includes a path

guillomovitch:
This PR fixes issue #179. 

I didn't fix metadata endpoint registration, tough.


bajnokk:
Note there is a related PR #451 , which tries to take care of all backend and frontend modules. I'd be happy if you could point out, if there are key differences.

guillomovitch:
Thanks for the remark, I didn't noticed this PR. It seems we have  now three different issues (#148, #179, #404), and two different PR (#451, #483) about the same problem :)

Regarding your own PR, it seems far more complete than mine, which was a quick'n'dirty fix for my own use case. I lack knowledge of SATOSA internals, but if I understand correctly, you modified routing to fix the issue. This strategies let persists inconsistencies between modules configurations, some using absolute URLs (SAML backend and frontends), ie:
<code>
        endpoints:
          assertion_consumer_service:
          - [<base_url>/<name>/acs/post, 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST']
</code>
while others (idpy-oidc fronted) using relatives one, ie:
<code>
endpoint:
        provider_info:
          path: .well-known/openid-configuration
</code>

I'd personally prefer to enforce the same overall convention for configuration files. And I also feel easier to deduce absolute URLs by appending a relative one to base URL, when needed, than to try to deduce relative URLs from absolute ones, using complex parsing.

BTW, your PR doesn't cover the SAML2 backend, which has to be fixed as well, at least for the reload-metada endpoint, and probably for others too.

